PR #7, opened 2017-03-01
Implemented a method to actually take a JSON payload and then do some…

sholaday:
…thing with it. Right now it just prints it as a placeholder.


schmidtw:
Based on our discussion earlier, should I still merge this?

sholaday:
No, I will continue on this branch with the new stuff that we talked about.

PR #9, opened 2017-03-10
Don't merge, just want to get some feedback.

schmidtw:
First pass at the outbound delivery mechanism.

### DIFF
--- a/src/caduceus/outboundSender.go
+++ b/src/caduceus/outboundSender.go
@@ -5,8 +5,10 @@ import (
 	"crypto/hmac"
 	"crypto/sha1"
 	"encoding/hex"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/Comcast/webpa-common/logging"
 	"hash"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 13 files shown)


src/caduceus/outboundSender_test.go
```diff
@@ -0,0 +1,263 @@
+package main
```
sholaday: Only real comment I have to make on the tests is that you could use this https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks (subtests) to have a bit less redundant set up and tear down code if you find you're having issues with that.

For example if I want to test the `New()` and `Destroy()` methods on my `HelloWorld` object:
```
func TestHelloWorld(t *testing.T) {
    // <setup code>
    t.Run("TestHelloWorldNew", func(t *testing.T) { ... })
    t.Run("TestHelloWorldDestroy", func(t *testing.T) { ... })
    // <tear-down code>
}
```

sholaday: Another thing I was going to mention was that you're not really using any interfaces in `outboundSender.go` but I guess you don't need to worry about that unless you needed to do some dependency injections for your unit tests.

src/caduceus/outboundSender.go
```diff
@@ -0,0 +1,221 @@
...
+	secret       []byte
+	events       []*regexp.Regexp
+	matcher      map[string][]*regexp.Regexp
+	queueSize    int
+	queue        chan OutboundRequest
+	wg           sync.WaitGroup
+}
+
+func (osf OutboundSenderFactory) New() (obs *OutboundSender, err error) {
+	if _, err = url.ParseRequestURI(osf.Url); nil != err {
```
schmidtw: Yep, no need trying to deliver to an invalid url.

src/caduceus/outboundSender.go
```diff
@@ -0,0 +1,221 @@
...
+	Events      []string
+	Matchers    map[string][]string
+	NumWorkers  int
+	QueueSize   int
+}
+
+type OutboundSender struct {
+	url          string
+	contentType  string
+	deliverUntil int64
```
schmidtw: Good idea - I'll see if I can incorporate this.

src/caduceus/outboundSender.go
```diff
@@ -0,0 +1,221 @@
...
+		queueSize:    osf.QueueSize,
+	}
+
+	if "" != osf.Secret {
+		tmp.secret = []byte(osf.Secret)
+	}
+
+	// Give us some head room so that we don't block when we get near the
+	// completely full point.
+	tmp.queue = make(chan OutboundRequest, osf.QueueSize+10)
```
schmidtw: yeah, this is ugly - my goal was to be able to try to not block the caller & force dump the channel.  That way if all the workers servicing this queue get busy, we can not back up the system for maybe 30 seconds.  I'll see if there's a better way to do this.

src/caduceus/outboundSender.go
```diff
@@ -0,0 +1,221 @@
...
+
+	tmp := &OutboundSender{
+		url:          osf.Url,
+		contentType:  osf.ContentType,
+		client:       osf.Client,
+		deliverUntil: osf.Until,
+		queueSize:    osf.QueueSize,
+	}
+
+	if "" != osf.Secret {
```
schmidtw: `if osf.Secret != "" {` is only 1 character away from `if osf.Secret = "" {`  Backwards will let the compiler throw an error, but is a bit different to look at :).

src/caduceus/outboundSender.go
```diff
@@ -0,0 +1,221 @@
...
+		tmp.secret = []byte(osf.Secret)
+	}
+
+	// Give us some head room so that we don't block when we get near the
+	// completely full point.
+	tmp.queue = make(chan OutboundRequest, osf.QueueSize+10)
+
+	// Create the event regex objects
+	for _, event := range osf.Events {
+		re, e := regexp.Compile(event)
```
schmidtw: 👍

PR #10, opened 2017-03-11
Feature/add health stats

sholaday:
Please review and contact me before merging. Going to ask John to glance over this.

### DIFF
--- a/src/caduceus/caduceus_test.go
+++ b/src/caduceus/caduceus_test.go
@@ -1,15 +1,13 @@
 package main
 
 import (
-	"bytes"
 	"github.com/Comcast/webpa-common/health"
 	"github.com/Comcast/webpa-common/logging"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
-	"io/ioutil"
-	"net/http"
 	"net/http/httptest"
 	"os"
+	"strings"
 	"testing"
 )
 
@@ -56,18 +54,15 @@ func TestServeHTTP(t *testing.T) {
 		}.New(),
 	}
 
-	testServer := httptest.NewServer(serverWrapper)
-	defer testServer.Close()
+	req := httptest.NewRequest("POST", "localhost:8080", strings.NewReader("Test payload."))
+	req.Header.Set("Content-Type", "text/plain")
+	w := httptest.NewRecorder()
 
-	buf := bytes.NewBufferString("Test message.")
+	serverWrapper.ServeHTTP(w, req)
 
-	res, err := http.Post(testServer.URL, "text/plain", buf)
-	assert.Nil(err)
-	defer res.Body.Close()
+	resp := w.Result()
 
-	resMsg, err := ioutil.ReadAll(res.Body)
-	assert.Nil(err)
-	assert.Equal("Request placed on to queue.\n", string(resMsg))
+	assert.Equal(202, resp.StatusCode)
 
 	fakeHandler.AssertExpectations(t)
 	fakeHealth.AssertExpectations(t)
--- a/src/caduceus/caduceus_test.go
+++ b/src/caduceus/caduceus_test.go
@@ -54,7 +54,7 @@ func TestServeHTTP(t *testing.T) {
 		}.New(),
 	}
 
-	t.Run("Test202", func(t *testing.T) {
+	t.Run("TestHappyPath", func(t *testing.T) {
 		req := httptest.NewRequest("POST", "localhost:8080", strings.NewReader("Test payload."))
 		req.Header.Set("Content-Type", "text/plain")
 		w := httptest.NewRecorder()
--- a/src/caduceus/caduceus_test.go
+++ b/src/caduceus/caduceus_test.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"bytes"
-	"fmt"
 	"github.com/Comcast/webpa-common/health"
 	"github.com/Comcast/webpa-common/logging"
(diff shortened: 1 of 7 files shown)


src/caduceus/caduceus_test.go
```diff
@@ -0,0 +1,197 @@
...
+	t.Run("TestWorkerPoolFullQueue", func(t *testing.T) {
+		err := workerPool.Send(func(workerID int) {
+			// do nothing
+		})
+
+		assert.NotNil(err)
+	})
+}
+
+func TestIncrementBucket(t *testing.T) {
```
johnabass: Change this to TestCaduceusHealth.  The idea is that the test name should refer to the type that's being tested followed by method or scenario.

src/caduceus/caduceus_test.go
```diff
@@ -0,0 +1,197 @@
...
+	assert := assert.New(t)
+
+	testData := []struct {
+		inSize       int
+		expectedStat health.Stat
+	}{
+		{inSize: 99, expectedStat: PayloadsOverZero},
```
johnabass: You really want more values here.  It's a good idea to check a negative value, zero, and extremes like math.MaxInt32 or math.MinInt32.  The goal of that is to make sure that future changes to the code being tested cause test breaks where they should.  For example, what if someone adds a PayloadsOverFifty stat?

You can't be exhaustive, of course.  But spot checking more values makes the test better able to warn a maintenance programmer that he's done something that breaks your assumptions.

src/caduceus/caduceus_test.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func (m *mockHealthTracker) ServeHTTP(response http.ResponseWriter, request *http.Request) {
+	m.Called(response, request)
+	return
+}
+
+func (m *mockHealthTracker) IncrementBucket(inSize int) {
+	m.Called(inSize)
+	return
```
johnabass: You don't need a return here, fyi.

src/caduceus/caduceus_test.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+func (m *mockHealthTracker) SendEvent(healthFunc health.HealthFunc) {
+	m.Called(healthFunc)
+	return
+}
+
+func (m *mockHealthTracker) ServeHTTP(response http.ResponseWriter, request *http.Request) {
+	m.Called(response, request)
+	return
```
johnabass: You don't need a return here, fyi.

src/caduceus/caduceus_test.go
```diff
@@ -0,0 +1,197 @@
...
+
+// mockHealthTracker needs to mock things from both the `HealthTracker`
+// interface as well as the `health.Monitor` interface
+type mockHealthTracker struct {
+	mock.Mock
+}
+
+func (m *mockHealthTracker) SendEvent(healthFunc health.HealthFunc) {
+	m.Called(healthFunc)
+	return
```
johnabass: You don't need a return here, fyi.

src/caduceus/caduceus_test.go
```diff
@@ -0,0 +1,197 @@
...
+// Begin mock declarations
+
+// mockHandler only needs to mock the `HandleRequest` method
+type mockHandler struct {
+	mock.Mock
+}
+
+func (m *mockHandler) HandleRequest(workerID int, inRequest CaduceusRequest) {
+	m.Called(workerID, inRequest)
+	return
```
johnabass: You don't need a return here, fyi.

src/caduceus/caduceus_test.go
```diff
@@ -0,0 +1,197 @@
...
+			caduceusHealth := &CaduceusHealth{fakeMonitor}
+
+			caduceusHealth.IncrementBucket(data.inSize)
+			fakeMonitor.AssertExpectations(t)
+		}
+	})
+}
+
+func TestServeHTTP(t *testing.T) {
+	assert := assert.New(t)
```
johnabass: Test names should refer to a type first, not a method.  TestFooServeHTTP would be more consistent with golang style, where Foo was a type that had a ServeHTTP method.  Alternatively, call this TestFoo and use the new t.Run(...) syntax for each method or scenario of Foo.

src/caduceus/caduceus_test.go
```diff
@@ -0,0 +1,197 @@
...
+func TestWorkerPool(t *testing.T) {
+	assert := assert.New(t)
+
+	workerPool := WorkerPoolFactory{
+		NumWorkers: 1,
+		QueueSize:  1,
+	}.New()
+
+	t.Run("TestWorkerPoolCreation", func(t *testing.T) {
+		assert.NotNil(workerPool)
```
johnabass: This subtest really doesn't do much.  If you want to specify that a condition is a requirement for the test to proceed, as opposed to just asserting state, use the github.com/stretchr/testify/require API.

src/caduceus/caduceus_test.go
```diff
@@ -0,0 +1,197 @@
...
+	})
+
+	workerPool = WorkerPoolFactory{
+		NumWorkers: 1,
+		QueueSize:  0,
+	}.New()
+
+	t.Run("TestWorkerPoolFullQueue", func(t *testing.T) {
+		err := workerPool.Send(func(workerID int) {
+			// do nothing
```
johnabass: This function should assert that it got executed.  Also, this subtest should probably block and wait for that assertion.  You can use a sync.WaitGroup or just a simple `chan struct{}` to block.

PR #34, opened 2017-07-31
make requests to insure dns is ready before preceding.

njharter:


src/caduceus/caduceus_type.go
```diff
@@ -105,6 +106,23 @@ func (ch *CaduceusHealth) IncrementBucket(inSize int) {
 	}
 }
 
+func (ch *CaduceusHealth) dnsReady(address string, ready chan bool) {	
+	req, err := http.NewRequest("GET", address, nil)
+	if err != nil {
+		return
+	}
+	client := http.Client{}
```
shilpa24balaji: Is it not better to do connection reuse?

PR #52, opened 2017-09-01
Make log strings more readable as structured ones.

joe94:


src/caduceus/http.go
```diff
@@ -35,16 +37,16 @@ func (sh *ServerHandler) ServeHTTP(response http.ResponseWriter, request *http.R
 		response.WriteHeader(http.StatusBadRequest)
 		response.Write([]byte(fmt.Sprintf("Unsupported method \"%s\"... Caduceus only supports \"POST\" method.\n", request.Method)))
 
-		debugLog.Log(messageKey, "Unsupported method", "method", request.Method, "explanation",
-		"Caduceus only supports \"POST\" method.")
+		errorLog.Log(errorKey, "Unsupported method", "method", request.Method, messageKey,
```
joe94: @johnabass This log did not seem a debug level (as it was before switching to go-kit) one but rather that of an error. Correct if I am wrong.

src/caduceus/http.go
```diff
@@ -35,16 +37,16 @@ func (sh *ServerHandler) ServeHTTP(response http.ResponseWriter, request *http.R
...
+		"Caduceus only supports POST method.")
 		return
 	}
 
 	payload, err := ioutil.ReadAll(request.Body)
 	if err != nil {
 		response.WriteHeader(http.StatusBadRequest)
 		response.Write([]byte(fmt.Sprintf("Unable to retrieve the request body: %s.\n", err.Error)))
-		debugLog.Log(messageKey, "Unable to retrieve the request body.", logging.ErrorKey(), err.Error)
+		errorLog.Log(messageKey, "Unable to retrieve the request body.", errorKey, err.Error)
```
joe94: @johnabass same as above.

src/caduceus/http.go
```diff
@@ -62,8 +64,8 @@ func (sh *ServerHandler) ServeHTTP(response http.ResponseWriter, request *http.R
 				response.Write([]byte(fmt.Sprintf("Only Content-Type values of \"application/json\" or " +
 					"\"application/msgpack\" are supported got: [%s].\n", value)))
 
-				debugLog.Log("explanation", "explanation","Only Content-Type values of \"application/json\" or " +
-					"\"application/msgpack\" are supported.", "actual", value)
+				errorLog.Log(messageKey,"Only Content-Type values 'application/json' or 'application/msgpack' are " +
```
joe94: Same ^

src/caduceus/outboundSender.go
```diff
@@ -503,7 +500,7 @@ func (obs *CaduceusOutboundSender) queueOverflow() {
 						"notification", obs.listener.FailureURL)
 				} else {
 					// Success
-					errorLog.Log("Able to send cut-off notification", "url", obs.listener.FailureURL,
+					logging.Info(obs.logger).Log("Able to send cut-off notification", "url", obs.listener.FailureURL,
```
joe94: @johnabass It seemed odd that a success case was being logged at error level.

src/caduceus/http.go
johnabass: This log should be removed entirely.  The mapping for the handler should prevent unsupported methods up front, so code should never get into this case.

src/caduceus/http.go
johnabass: This log works fine, but code should not attempt to write anything to the response.  If a request body occurs, that indicates a low-level socket I/O error and the response/request pair should be abandoned.

src/caduceus/http.go
johnabass: Again, the mapping should prevent content types that are unsupported from reaching this block of code.  Remove the log and verify that there's a `Headers("Content-Type", ...)` mapper.

src/caduceus/http.go
```diff
@@ -118,6 +120,7 @@ type ProfileHandler struct {
 // that the main handler has successfully dealt with
 func (ph *ProfileHandler) ServeHTTP(response http.ResponseWriter, request *http.Request) {
 	logging.Info(ph.Logger).Log(logging.MessageKey(), "Receiving request for server stats...")
```
johnabass: This is way too much logging.  Remove this entirely.  If we want request logs, that should be handled in middleware.

src/caduceus/senderWrapper_test.go
```diff
@@ -46,7 +47,7 @@ func (t *swTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 }
 
 func swGetLogger() log.Logger {
```
johnabass: This function is a meaningless abstraction.  Just replace calls to `swGetLogger` to `logging.DefaultLogger()`

src/caduceus/http.go
joe94: Understood. Removed lines 47-48.

src/caduceus/http.go
joe94: Addressed.

src/caduceus/http.go
joe94: Addressed.

PR #73, opened 2018-02-21
add dep files

joe94:
@johnabass I could not update the webpacommon version on the glide files. Do you mind helping with those?


schmidtw:
We're not going to move to dep yet.

PR #87, opened 2018-06-18
The tests inside senderWrapper_test need to be rethought since they h…

schmidtw:
…appen asynchronously.


schmidtw:
This is accumulating technical debt.  The reason this code doesn't work is because the timeout in webpa common is hard coded to 5min at this point.

PR #95, opened 2018-08-03
removed sending duplicate events to the same webhook

kcajmagic:


njharter:
please update CHANGELOG.md

PR #96, opened 2018-08-07
fixed logic for updating webhooks

kcajmagic:
- webhooks will no longer append regex to previous webhooks
  (aka matcher and events will start from scratch)


njharter:
please update CHANGELOG.md

PR #106, opened 2018-10-23
added consul registration

kcajmagic:
Will be used by talaria for round robin load balancing.


schmidtw:
Will this automatically mark the Caduceus service in Consul as `passing`, `warning`, `critical` based on the Caduceus health?

PR #112, opened 2018-11-16
Modified Workers to stop or start based on success rate

kristinaspring:


schmidtw:
With the change to no worker pool I don't think we need this pr.

PR #119, opened 2019-02-16
Remove the outbound worker pool and replace it with short lived gorou…

schmidtw:
…tines and a maximum limit configuration.


src/caduceus/outboundSender.go
```diff
@@ -343,70 +343,83 @@ func (obs *CaduceusOutboundSender) Queue(msg *wrp.Message) {
...
-			obs.droppedExpiredCounter.Add(1.0)
+		*/
+		if matchDevice {
+			if len(obs.queue) < obs.queueSize {
+				obs.queueDepthGauge.Add(1.0)
+				obs.queue <- msg
+				debugLog.Log(logging.MessageKey(), "WRP Sent to obs queue", "url", obs.id)
+				// a regex was matched, no need to check further matches
+				break
+			} else {
```
kristinaspring: this else isn't necessary.

schmidtw: I cleaned up the ones in outboundSender.go.  I think it makes sense to do the other files in a different PR.

kristinaspring: this is outboundSender.go, but more cleanup in a different PR sounds good to me.

Issue #117, opened 2019-02-11
Add event type metric for incoming events.


schmidtw:
Caduceus should provide an incoming (versus existing delivered) metric for count by event type.  This will help us understand all the events vs just the ones being delivered.

PR #131, opened 2019-03-11
caduceus: add metric - all event types before dropped

githubsands:
Solves #117 and moves caduceus to wrp-go.

### DIFF
--- a/src/caduceus/metrics.go
+++ b/src/caduceus/metrics.go
@@ -13,6 +13,7 @@ const (
 	SlowConsumerCounter           = "slow_consumer_cut_off_count"
 	IncomingQueueDepth            = "incoming_queue_depth"
 	IncomingContentTypeCounter    = "incoming_content_type_count"
+	IncomingEventTypeCounter      = "incoming_event_type_count"
 	DropsDueToInvalidPayload      = "drops_due_to_invalid_payload"
 	OutgoingQueueDepth            = "outgoing_queue_depths"
 )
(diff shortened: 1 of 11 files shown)


src/caduceus/senderWrapper.go
```diff
@@ -163,6 +168,8 @@ func (sw *CaduceusSenderWrapper) Update(list []webhook.W) {
 // function performs the fan-out and filtering to multiple possible endpoints.
 func (sw *CaduceusSenderWrapper) Queue(msg *wrp.Message) {
 	sw.mutex.RLock()
+	sw.eventType.With("event_type", strings.TrimLeft(msg.Destination[23:], "/")).Add(1.0)
```
schmidtw: This should ideally be done by WRP library... but until then, we should do it the same way as https://github.com/Comcast/caduceus/blob/master/src/caduceus/outboundSender.go#L514

src/caduceus/metrics.go
```diff
@@ -75,5 +76,11 @@ func Metrics() []xmetrics.Metric {
 			Type:       "gauge",
 			LabelNames: []string{"url"},
 		},
+		{
+			Name:       IncomingEventTypeCounter,
+			Help:       "Incoming count of events by event type",
+			Type:       "counter",
+			LabelNames: []string{"event_type"},
```
schmidtw: `event_type` should be `event` to be consistent with other metrics.

src/caduceus/senderWrapper.go
```diff
@@ -163,6 +167,9 @@ func (sw *CaduceusSenderWrapper) Update(list []webhook.W) {
 // function performs the fan-out and filtering to multiple possible endpoints.
 func (sw *CaduceusSenderWrapper) Queue(msg *wrp.Message) {
 	sw.mutex.RLock()
+
+	sw.eventType.With("event_type", msg.FindEventStringSubMatch())
```
schmidtw: Changes to `event`.

PR #133, opened 2019-03-12
caduceus: add metric - incoming content type

githubsands:
This implements IncomingContentType metric. It was within a caduceus object but unimplemented.


src/caduceus/outboundSender_test.go
```diff
@@ -122,18 +122,30 @@ func simpleFactorySetup(trans *transport, cutOffPeriod time.Duration, matcher []
 	fakeDroppedSlow.On("With", []string{"url", w.Config.URL, "reason", "network_err"}).Return(fakeDroppedSlow)
 	fakeDroppedSlow.On("Add", 1.0).Return()
 
-	// test queue depth
+	// IncomingContentType cases
+	fakeContentType := new(mockCounter)
+	fakeContentType.On("With", []string{"content", "msgpack"}).Return(fakeContentType)
```
schmidtw: Do these need to be changed to be `content_type` as well?

PR #134, opened 2019-03-14
caduceus: migrate to wrp-go from webpa-common/wrp

githubsands:
Changes webpa-common/wrp dependencies to wrp-go/wrp and also updates glide.


githubsands:
solved by incoming event type pr

PR #137, opened 2019-05-10
Make a first pass at documenting the parameters available in the conf…

schmidtw:
…iguration yaml file.


schmidtw:
This needs to be tested to make sure the yaml will work before mergins.

PR #147, opened 2019-06-26
Fix the race condition and make the alternate urls automatically roun…

schmidtw:
…d-robin along the way.

### DIFF
--- a/src/caduceus/outboundSender.go
+++ b/src/caduceus/outboundSender.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"bytes"
+	"container/ring"
 	"crypto/hmac"
 	"crypto/sha1"
 	"encoding/hex"
@@ -114,6 +115,7 @@ type OutboundSender interface {
 // CaduceusOutboundSender is the outbound sender object.
 type CaduceusOutboundSender struct {
 	id                       string
+	urls                     *ring.Ring
 	listener                 webhook.W
 	deliverUntil             time.Time
 	dropUntil                time.Time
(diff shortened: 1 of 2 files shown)


src/caduceus/outboundSender.go
```diff
@@ -442,7 +471,7 @@ func (obs *CaduceusOutboundSender) dispatcher() {
 
 // worker is the routine that actually takes the queued messages and delivers
 // them to the listeners outside webpa
-func (obs *CaduceusOutboundSender) send(secret, acceptType string, msg *wrp.Message) {
+func (obs *CaduceusOutboundSender) send(urls *ring.Ring, secret, acceptType string, msg *wrp.Message) {
```
kcajmagic: `urls` is part of the `CaduceusOutboundSender`. We don't need to change the method signature.

src/caduceus/outboundSender.go
```diff
@@ -465,12 +494,12 @@ func (obs *CaduceusOutboundSender) send(secret, acceptType string, msg *wrp.Mess
 	}
 	payloadReader = bytes.NewReader(body)
 
-	req, err := http.NewRequest("POST", obs.id, payloadReader)
+	req, err := http.NewRequest("POST", urls.Value.(string), payloadReader)
```
kcajmagic: For the first request, we should use the primary url (aka `id`). If that fails then we can leverage the alternative urls.
```suggestion
	req, err := http.NewRequest("POST", obs.id, payloadReader)
```

src/caduceus/outboundSender.go
```diff
@@ -412,6 +437,10 @@ func (obs *CaduceusOutboundSender) dispatcher() {
 		obs.queueDepthGauge.Add(-1.0)
 
 		obs.mutex.RLock()
+		urls := obs.urls
+		// Move to the next URL to try 1st the next time.
+		obs.urls = obs.urls.Next()
```
kcajmagic: I don't think this is needed.

src/caduceus/outboundSender.go
schmidtw: By sending the urls, we avoid the race condition where the `CaduceusOutboundSender` version could change, or needing to lock.  We don't want to lock if at all possible.

src/caduceus/outboundSender.go
schmidtw: This allows us to move on to the next url & load balance across all alternatives.

schmidtw:
The nice thing about leveraging the alternatives for mainstream delivery, we provide our customers a mechanism for very uniform load balancing if they so desire.  I am thinking we just use the base url or the alternatives to make it simpler for callers of the api.  We'd still use the base url for the id in both cases.

PR #149, opened 2019-07-15
fix dockerfiles

joe94:


joe94:
Test `TestSwSimple` does not work with with the go test `-race` flag.

joe94:
I don't see the error on this branch anymore: https://github.com/xmidt-org/caduceus/pull/147

Issue #151, opened 2019-08-06
Caduceus panics with non-utf string for event name


ilawjr:
If you attempt to pass in a string that cannot be parsed correctly by the underlying metrics, caduceus panics.  

panic: label event: value "raw.kestrel.reports\xedNetworkDevicesStatus" is not valid UTF-8

vendor/github.com/prometheus/client_golang/prometheus.(*CounterVec).With(0xc00000ea08, 0xc05d2c21b0, 0x6, 0xc05d2c21b0)
    /root/rpmbuild/BUILD/caduceus-0.1.3/src/vendor/github.com/prometheus/client_golang/prometheus/counter.go:217 +0x79
vendor/github.com/go-kit/kit/metrics/prometheus.(*Counter).Add(0xc010491dc0, 0x3ff0000000000000)
    /root/rpmbuild/BUILD/caduceus-0.1.3/src/vendor/github.com/go-kit/kit/metrics/prometheus/prometheus.go:45 +0x5c
main.(*CaduceusOutboundSender).send(0xc001826900, 0xc01628c590, 0xe, 0xc01628c578, 0x4, 0xc0a0141b00)
    /root/rpmbuild/BUILD/caduceus-0.1.3/src/caduceus/outboundSender.go:537 +0xa44
created by main.(*CaduceusOutboundSender).dispatcher
    /root/rpmbuild/BUILD/caduceus-0.1.3/src/caduceus/outboundSender.go:444 +0x342

schmidtw:
This seems to be something that should be done in wrp-go based on some local feedback I got.

https://github.com/xmidt-org/wrp-go/issues/25

We also should make the outgoing goroutine panic tolerant as well.

kristinaspring:
This is fixed.

PR #152, opened 2019-08-06
Add a fix where a panic in the sender go routine doesn't kill the ent…

schmidtw:
…ire machine.

#151

### DIFF
--- a/src/caduceus/mocks_test.go
+++ b/src/caduceus/mocks_test.go
@@ -19,6 +19,7 @@ package main
 import (
 	"net/http"
 	"time"
+	"unicode/utf8"
 
 	"github.com/Comcast/webpa-common/health"
 	"github.com/Comcast/webpa-common/webhook"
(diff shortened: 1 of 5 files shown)


src/caduceus/mocks_test.go
```diff
@@ -99,6 +100,11 @@ func (m *mockCounter) Add(delta float64) {
 }
 
 func (m *mockCounter) With(labelValues ...string) metrics.Counter {
+	for _, v := range labelValues {
+		if !utf8.ValidString(v) {
+			panic("not UTF-8")
+		}
+	}
 	args := m.Called(labelValues)
```
kristinaspring: Can this line be before the labelValues check?

src/caduceus/mocks_test.go
```diff
@@ -117,6 +123,11 @@ func (m *mockGauge) Set(value float64) {
 }
 
 func (m *mockGauge) With(labelValues ...string) metrics.Gauge {
+	for _, v := range labelValues {
+		if !utf8.ValidString(v) {
+			panic("not UTF-8")
+		}
+	}
 	args := m.Called(labelValues)
```
kristinaspring: Can this line be before the labelValues check?

src/caduceus/mocks_test.go
schmidtw: yes - good catch

src/caduceus/mocks_test.go
schmidtw: yes - good catch

schmidtw: I flipped it back so we don't need to add non UTF-8 values to the tests... we will clean this up when we do some pretty serious house cleaning on the code.

Issue #155, opened 2019-08-26
Make annotated yaml easier to find


kristinaspring:
https://github.com/xmidt-org/caduceus/blob/58479b028a3c90cccd40bcde51fbc37e53718f6e/etc/caduceus/caduceus.yaml#L1

Can we put a link to this yaml in the README or otherwise make it easier to find in the repository for anyone with questions? There's a yaml in the root directory, but it doesn't have explanations so someone new to the repository may not find this annotated one.

kristinaspring:
Fixed in https://github.com/xmidt-org/caduceus/pull/158

PR #157, opened 2019-08-29
Update example location

schmidtw:
Fix GH Issue #155


kristinaspring:
This is fixed in https://github.com/xmidt-org/caduceus/pull/158

Issue #185, opened 2020-03-19
Caduceus cutoffs


gargidb:
Currently caduceus is able to send a cutoff message when already inside a cutoff window, as a result, we get multiple cutoffs. We want to prevent caduceus from sending another cutoff by checking if we are already inside a cutoff window.

Add check before dropUntil

[https://github.com/xmidt-org/caduceus/blob/d445e08bdecabe906bfcc1bb9506a860874cef92/outboundSender.go#L606](url)

joe94:
fixed by #186

PR #186, opened 2020-03-19
caduceus queue cutoff

gargidb:
[https://github.com/xmidt-org/caduceus/issues/185](url)


schmidtw:
Great find.  Thank you!

Issue #189, opened 2020-03-26
Idiomatic full queue check


kristinaspring:
The check here can be written differently, where if we can't immediately push the event onto the channel, we know the queue is full:
https://github.com/xmidt-org/caduceus/blob/master/outboundSender.go#L418

Here is an example of how this can be written in a more idiomatic way:
https://gobyexample.com/non-blocking-channel-operations

This was originally brought up by @joe94 .

Issue #188, opened 2020-03-26
Caduceus drop events efficiently before cutoff ends


gargidb:
Caduceus currently does not completely empty the queue before the cutoff period ends, as a result causing multiple cutoffs. 
Need to efficiently drop all events from queue before cutoff ends.

PR #190, opened 2020-03-27
Caduceus queue

gargidb:
[https://github.com/xmidt-org/caduceus/issues/188](url)
[https://github.com/xmidt-org/caduceus/issues/189](url)

### DIFF
--- a/outboundSender.go
+++ b/outboundSender.go
@@ -195,6 +195,7 @@ func (osf OutboundSenderFactory) New() (obs OutboundSender, err error) {
 			QueueSize:    osf.QueueSize,
 			Workers:      osf.NumWorkers,
 		},
+		queueEmpty: true,
 	}
 
 	// Don't share the secret with others when there is an error.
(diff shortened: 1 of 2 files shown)


outboundSender.go
```diff
@@ -415,13 +416,15 @@ func (obs *CaduceusOutboundSender) Queue(msg *wrp.Message) {
...
 		*/
 		if matchDevice {
-			if len(obs.queue) < obs.queueSize {
+			select {
+			case obs.queue <- msg:
 				obs.queueDepthGauge.Add(1.0)
-				obs.queue <- msg
 				debugLog.Log(logging.MessageKey(), "WRP Sent to obs queue", "url", obs.id)
-				// a regex was matched, no need to check further matches
 				break
```
kristinaspring: Unfortunately, because the `break` is in a select statement, it now applies to the `select` rather than the for loop it originally applied to:
https://golang.org/ref/spec#Break_statements
I would suggest just removing the break (and the continue), and moving lines 428 and 429 inside of the `default` option for the select statement.  Since there is nothing that needs to apply to both cases, we don't need anything outside of the `select`.

outboundSender.go
```diff
@@ -431,7 +434,7 @@ func (obs *CaduceusOutboundSender) Queue(msg *wrp.Message) {
 func (obs *CaduceusOutboundSender) isValidTimeWindow(now, dropUntil, deliverUntil time.Time) bool {
 	var debugLog = logging.Debug(obs.logger)
 
-	if false == now.After(dropUntil) {
+	if false == now.After(dropUntil) && obs.queueEmpty {
```
kristinaspring:
```suggestion
	if false == now.After(dropUntil) || !obs.queueEmpty {
```
we want to check for when we are before the drop until or that the queue isn't empty - as long as one of these things is true, the client is still cut off.

outboundSender.go
```diff
@@ -448,13 +451,29 @@ func (obs *CaduceusOutboundSender) isValidTimeWindow(now, dropUntil, deliverUnti
 	return true
 }
 
+func (obs *CaduceusOutboundSender) Empty() {
+	for obs.queueEmpty == false {
```
kristinaspring:
```suggestion
	for !obs.queueEmpty {
```
this is a superficial thing, feel free to ignore.

outboundSender.go
```diff
@@ -448,13 +451,29 @@ func (obs *CaduceusOutboundSender) isValidTimeWindow(now, dropUntil, deliverUnti
...
+}
+
 func (obs *CaduceusOutboundSender) dispatcher() {
 	defer obs.wg.Done()
 
 	for msg := range obs.queue {
 		obs.queueDepthGauge.Add(-1.0)
 
 		obs.mutex.RLock()
+		if obs.queueEmpty == false {
```
kristinaspring:
```suggestion
		if !obs.queueEmpty {
```
this is a superficial thing, feel free to ignore.

PR #202, opened 2020-05-08
switch webhook configuration from sns to argus

kcajmagic:


caduceus.yaml
```diff
@@ -343,4 +330,4 @@
   # perSourceIncomingPayloadSizeBuckets:
   #   - 100
   #   - 1000
-  #   - 10000
+  #   - 10000
```
joe94, line 333:
```suggestion
  #   - 10000
  
```
unix text file newline at eof

webhookHandler.go
```diff
@@ -0,0 +1,107 @@
...
+	} else {
+		rw.Header().Set("Content-Type", "application/json")
+		rw.Write(msg)
+	}
+}
+
+// update is an api call to processes a listenener registration for adding and updating
+func (r *Registry) UpdateRegistry(rw http.ResponseWriter, req *http.Request) {
+	payload, err := ioutil.ReadAll(req.Body)
+	req.Body.Close()
```
joe94, line 92:
```suggestion
```

The underlying already does this.

webhookHandler.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+type RegistryConfig struct {
+	Logger      log.Logger
+	Listener    webhookclient.ListenerFunc
+	ArgusConfig webhookclient.ClientConfig
+}
+
+func NewRegistry(config RegistryConfig, listener webhookclient.Listener) (*Registry, error) {
+	argus, err := webhookclient.CreateClient(config.ArgusConfig, webhookclient.WithLogger(config.Logger))
```
joe94, line 27: where are we registering the config.Listener (caduceusSenderWrapper.Update in this case)?

caduceus.yaml
```diff
@@ -230,59 +230,46 @@
...
-  # (Optional)
-  soa:
-    provider: "example.awsdns-19.com:53"
+  authHeader: ["xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx=","dXNlcjpwYXNz"]
 
 ########################################
 #   Webhook Related Configuration
 ########################################
+  # Argus Config for storing the webhook information
+  argusConfig:
```
joe94, line 239: Should we make the toplevel field `webhook` here and place all the webhook related configs under it? Then we can do the same for scytale and keep both consistent.

caduceus.yaml
```diff
@@ -230,59 +230,46 @@
...
-  # (Optional)
-  soa:
-    provider: "example.awsdns-19.com:53"
+  authHeader: ["xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx=","dXNlcjpwYXNz"]
 
 ########################################
 #   Webhook Related Configuration
 ########################################
+  # Argus Config for storing the webhook information
+  webhookConfig:
```
joe94, line 239: Why not just `webhook`?

main.go
```diff
@@ -138,33 +132,40 @@ func caduceus(arguments []string) int {
...
+		measures.WebhookListSize.Set(float64(len(items)))
 	}
-	webhookRegistry, webhookHandler := webhookFactory.NewRegistryAndHandler(metricsRegistry)
-	webhookFactory.SetExternalUpdate(caduceusSenderWrapper.Update)
+	webhookRegistry, err := NewRegistry(RegistryConfig{
+		Logger: logger,
+		Listener: func(items []model.Item) {
+			hooks := make([]webhook.W, len(items))
+			for index, item := range items {
+				hook, err := convertItemToWebhook(item)
```
joe94, line 146: if conversions fail, hooks will have holes with nil objects in them and the Update() function will panic https://github.com/xmidt-org/caduceus/blob/master/senderWrapper.go#L152
We should probably just append successfully converted objects and log error cases

caduceus.yaml
```diff
@@ -230,59 +230,46 @@
...
-  # can restart much faster.
-  # (Optional)
-  soa:
-    provider: "example.awsdns-19.com:53"
+  authHeader: ["xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx=","dXNlcjpwYXNz"]
 
 ########################################
 #   Webhook Related Configuration
 ########################################
+  # Argus Config for storing the webhook information
```
kristinaspring, line 238:
```suggestion
  # webhookStore provides configuration for storing and obtaining webhook 
  # information using argus.
```

caduceus_type.go
```diff
@@ -34,6 +35,7 @@ type CaduceusConfig struct {
 	JobQueueSize     int
 	Sender           SenderConfig
 	JWTValidators    []JWTValidator
+	WebhookConfig    chrysom.ClientConfig
```
kristinaspring, line 38:
```suggestion
	WebhookStore    chrysom.ClientConfig
```

deploy/packaging/caduceus_spruce.yaml
```diff
@@ -235,64 +235,18 @@ jwtValidators:
...
-    region: (( grab $AWS_REGION || "us-east-1" ))
-
-    # topicArn describes the SNS topic this server needs to subscribe to
-    topicArn: (( grab $AWS_SNS_ARN || "arn:aws:sns:us-east-1:000000000000:xmidt-local-caduceus"))
-
-    #urlPath is the URL path SNS will use to confirm a subscription with this server
-    urlPath: "/api/v2/aws/sns"
-
+# Argus Config for storing the webhook information
+webhookConfig:
```
kristinaspring, line 244:
```suggestion
webhookStore:
```

caduceus.yaml
kristinaspring, line 239:
```suggestion
  webhookStore:
```

main.go
```diff
@@ -82,13 +82,7 @@ func caduceus(arguments []string) int {
...
 	}
 
-	var (
-		infoLog  = log.WithPrefix(logger, level.Key(), level.InfoValue())
-		errorLog = log.WithPrefix(logger, level.Key(), level.ErrorValue())
-		debugLog = log.WithPrefix(logger, level.Key(), level.DebugValue())
-	)
-
-	infoLog.Log("configurationFile", v.ConfigFileUsed())
+	logging.Info(logger).Log("configurationFile", v.ConfigFileUsed())
```
kristinaspring, line 85: I just got rid of all these `logging.Info()` in caduceus to increase performance - please use
```
log.WithPrefix(logger, level.Key(), level.InfoValue()).Log(whatever...)
``` 
or add the level.Key() stuff to the log statement itself.

main.go
```diff
@@ -138,33 +132,40 @@ func caduceus(arguments []string) int {
...
-	webhookRegistry, webhookHandler := webhookFactory.NewRegistryAndHandler(metricsRegistry)
-	webhookFactory.SetExternalUpdate(caduceusSenderWrapper.Update)
+	webhookRegistry, err := NewRegistry(RegistryConfig{
+		Logger: logger,
+		Listener: func(items []model.Item) {
+			hooks := make([]webhook.W, len(items))
+			for index, item := range items {
+				hook, err := convertItemToWebhook(item)
+				if err != nil {
+					continue
```
kristinaspring, line 148: we should probably log this or put it in a metric or something - we don't expect any failures here and suppressing this error could make it more difficult to debug later.

main.go
```diff
@@ -203,23 +191,7 @@ func caduceus(arguments []string) int {
...
-		errorLog.Log(logging.ErrorKey(), webhookStartResults.Error)
-	} else {
-		// todo: add message
-		webhookFactory.SetList(webhook.NewList(webhookStartResults.Hooks))
-		caduceusSenderWrapper.Update(webhookStartResults.Hooks)
-	}
-
-	debugLog.Log(messageKey, "Current listener retrieval.", "elapsedTime", time.Since(beginObtainList))
-	infoLog.Log(messageKey, "Caduceus is up and running!", "elapsedTime", time.Since(beginCaduceus))
+	logging.Info(logger).Log(logging.MessageKey(), "Caduceus is up and running!", "elapsedTime", time.Since(beginCaduceus))
```
kristinaspring, line 194: This also shouldn't use `logging.Info()`.

webhookHandler.go
```diff
@@ -0,0 +1,141 @@
...
+
+// jsonResponse is an internal convenience function to write a json response
+func jsonResponse(rw http.ResponseWriter, code int, msg string) {
+	rw.Header().Set("Content-Type", "application/json")
+	rw.WriteHeader(code)
+	rw.Write([]byte(fmt.Sprintf(`{"message":"%s"}`, msg)))
+}
+
+// get is an api call to return all the registered listeners
+func (r *Registry) GetRegistry(rw http.ResponseWriter, req *http.Request) {
```
kristinaspring, line 50: This is a change to the Caduceus API - previously, we were returning the registry that Caduceus has stored locally.  Now, we are returning the list of webhooks that Argus gives us, but not the list that Caduceus is currently working with.  A Get() on chrysom also doesn't even update the listener, so Caduceus could actually return a list with webhooks it doesn't have yet.

Issue #208, opened 2020-05-13
Exiting dispatcher() on a cutoff


kristinaspring:
Once we cut a webhook off, the long running goroutine dispatcher() exits.  Until this outbound sender goes away and we make a new one for the webhook, they'll never get another message from us.

https://github.com/xmidt-org/caduceus/blob/b887423704e38e13c8f91a0d5d24af3d157e4020/outboundSender.go#L487

We shouldn't break here; instead there should be a `continue`.

PR #210, opened 2020-05-13
fixed dispatcher() on cutoff conditions

kristinaspring:
fixes https://github.com/xmidt-org/caduceus/issues/208

### DIFF
--- a/outboundSender.go
+++ b/outboundSender.go
@@ -440,6 +440,10 @@ func (obs *CaduceusOutboundSender) isValidTimeWindow(now, dropUntil, deliverUnti
 	return true
 }
 
+// Empty is called on cutoff or shutdown and swaps out the current queue for
+// a fresh one, counting any current messages in the queue as dropped.
+// It should never close a queue, as a queue not referenced anywhere will be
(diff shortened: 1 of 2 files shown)


schmidtw:
Remove the `Loop:` label.

kristinaspring:
We need `Loop:` for this line though:
https://github.com/xmidt-org/caduceus/blob/b887423704e38e13c8f91a0d5d24af3d157e4020/outboundSender.go#L470

Otherwise, we would just `break` out of the select statement.  Is there a better way to do this?

outboundSender.go
```diff
@@ -480,7 +480,7 @@ Loop:
 
 			if now.Before(dropUntil) {
 				obs.droppedCutoffCounter.Add(1.0)
-				break Loop
+				continue
```
johnabass, line 483: This works, but do we still need the `Loop` label at all?

Quick thought:  Seems like this code never exits now, which isn't much of an issue if we never intend to quiesce/restart it.

gargidb:
I think we still need the `Loop` label to break out if the queue is empty.

Issue #222, opened 2020-06-12
Device isn't always the Source in the WRP


kristinaspring:
https://github.com/xmidt-org/caduceus/blob/b2866f45a7891e20b49c005aaf46dd70baf247e8/outboundSender.go#L391

Now that Talaria creates a few wrp simple events, this check could filter them out if a webhook is registered for specific devices.

kristinaspring:
There is a similar issue on this line:
https://github.com/xmidt-org/caduceus/blob/453f1410b1f109beeac7d9141eb20df4c86292d6/outboundSender.go#L582

kristinaspring:
Closing this in favor of the changes from the discussion related to https://github.com/xmidt-org/caduceus/issues/229

PR #228, opened 2020-07-28
added configurable device id parsing

kristinaspring:
fixes https://github.com/xmidt-org/caduceus/issues/222

uses the new wrp-listener library:
https://github.com/xmidt-org/wrp-listener/tree/main/wrpParser


schmidtw:
@kristinaspring and I talked about the changes & we think the changes bring to light a few important issues that she highlights in issue #229 .  I don't think Caduceus should be in the business of needing to do "smart parsing" of the wrp locators used as the wrp.source / dest fields to figure out the device id.  It unfortunately has stepped down that path, but we should reconsider if there is a better way to achieve the desired outcome versus this approach.

PR #235, opened 2020-09-24
fixed chrysom client not starting

kcajmagic:


main.go
```diff
@@ -135,27 +136,26 @@ func caduceus(arguments []string) int {
 	}
 	measures := NewMeasures(metricsRegistry)
 
-	var updateListSizeMetric chrysom.ListenerFunc
-	updateListSizeMetric = func(items []model.Item) {
+	caduceusConfig.WebhookStore.Logger = logger
+	caduceusConfig.WebhookStore.HTTPClient = http.DefaultClient
+	var webhookListener chrysom.ListenerFunc = func(items []model.Item) {
```
kristinaspring, line 141: Can you declare this function elsewhere and then use it here? Ideally we should have unit tests to cover this, right? And it seems impossible to unit test in its current state.

kristinaspring (changes requested):

kristinaspring (approved):

PR #237, opened 2020-10-07
Empty queue

### LINKED ISSUE #231: Empty queue when a webhook expires
https://github.com/xmidt-org/caduceus/blob/aa1419260f195f95b14c6534f40534c453c684ab/outboundSender.go#L517

Instead of allowing the dispatcher to clear the queue one event at a time when a webhook expires, call Empty() function to clear it.

j-mai:
When a webhook expires, empty the queue all at once instead of emptying one by one

Closes #231


outboundSender_test.go
```diff
@@ -918,3 +919,70 @@ func TestOverflow(t *testing.T) {
 
 	assert.NotNil(output.String())
 }
+
+func TestDispatcherEmptyQueue(t *testing.T) {
+	fmt.Printf("\n\nTestingDispatcherEmptyQueue:\n\n")
```
kcajmagic, line 924:
```suggestion
```
Adding this is unnecessary as we can run go test  -v to get which test it was apart of

kcajmagic (approved):

PR #263, opened 2021-04-26
Opentelemetry changes with optional tracing feature.

utsavbatra5:


http.go
```diff
@@ -43,9 +43,10 @@ type ServerHandler struct {
 }
 
 func (sh *ServerHandler) ServeHTTP(response http.ResponseWriter, request *http.Request) {
-	debugLog := log.WithPrefix(sh.Logger, level.Key(), level.DebugValue())
-	infoLog := log.WithPrefix(sh.Logger, level.Key(), level.InfoValue())
-	errorLog := log.WithPrefix(sh.Logger, level.Key(), level.ErrorValue())
+	logger := log.With(logging.GetLogger(request.Context()))
```
joe94, line 46: As a defense mechanism, it might be good to fall back here on `sh.Logger` if `logging.GetLogger(...)` results in a nil logger.

http.go
```diff
@@ -43,9 +43,10 @@ type ServerHandler struct {
...
+	debugLog := log.With(logger, level.Key(), level.DebugValue())
+	infoLog := log.With(logger, level.Key(), level.InfoValue())
+	errorLog := log.With(logger, level.Key(), level.ErrorValue())
```
joe94, line 49: I believe we can use `github.com/go-kit/kit/log/level` here so that 
line 47 can simply be: 
```
debugLog := level.Debug(logger)
```

joe94 (changes requested):
Looks good so far! I'll leave a comment below with some missing cases.

joe94:
These are the missing cases I mention above:
1) Instrument Argus client: https://github.com/xmidt-org/tr1d1um/blob/8ab7921143ac368d484ee0c86fdf04fa98fe0850/main.go#L177

2) Instrument the sender client?? Do you have a use case for instrumenting the requests sending events to receivers?  If so, we'll want to instrument this transport https://github.com/xmidt-org/caduceus/blob/4360bd45390b519377058cd019d44ed8cfed6f7f/main.go#L94
Example: https://github.com/xmidt-org/tr1d1um/blob/8ab7921143ac368d484ee0c86fdf04fa98fe0850/main.go#L370
This will be especially helpful if you instrument your event receivers as well.

main.go
```diff
@@ -63,6 +88,7 @@ func caduceus(arguments []string) int {
 		v = viper.New()
 
 		logger, metricsRegistry, webPA, err = server.Initialize(applicationName, arguments, f, v, Metrics, ancla.Metrics)
+		infoLogger                          = logging.Info(logger)
```
Sachin4403, line 91: @utsavbatra5  Here also we can use level.Info

utsavbatra5:
Hi @joe94, I have addressed the missing scenarios now with the new commit. Can you please review?

go.mod
```diff
@@ -2,20 +2,27 @@ module github.com/xmidt-org/caduceus
 
 go 1.14
 
+//replace github.com/xmidt-org/ancla => /Users/utsavbatra/Documents/work/ancla
+
```
joe94, line 6:
```suggestion
```

http.go
```diff
@@ -43,9 +43,16 @@ type ServerHandler struct {
...
 func (sh *ServerHandler) ServeHTTP(response http.ResponseWriter, request *http.Request) {
-	debugLog := log.WithPrefix(sh.Logger, level.Key(), level.DebugValue())
-	infoLog := log.WithPrefix(sh.Logger, level.Key(), level.InfoValue())
-	errorLog := log.WithPrefix(sh.Logger, level.Key(), level.ErrorValue())
+	var logger log.Logger
+	if logging.GetLogger(request.Context()) != nil {
+		logger = log.With(logging.GetLogger(request.Context()))
+	} else {
+		logger = sh.Logger
+	}
```
joe94, line 51:
```suggestion
	logger := logging.GetLogger(request.Context())
	if logger == nil {
	    logger = sh.Logger
	}
```

main.go
```diff
@@ -52,9 +57,78 @@ var (
...
 )
 
+// httpClientTimeout contains timeouts for an HTTP client and its requests.
+type httpClientTimeout struct {
+	// ClientTimeout is HTTP Client Timeout.
+	ClientTimeout time.Duration
+
+	// RequestTimeout can be imposed as an additional timeout on the request
+	// using context cancellation.
+	RequestTimeout time.Duration
```
joe94, line 67:
```suggestion
```

We're not doing explicit context-based cancellations here.

deploy/packaging/caduceus_spruce.yaml
```diff
@@ -338,3 +338,18 @@ sender:
...
+  # provider is the provider name. Currently, stdout, jaegar and zipkin are supported.
+  # 'noop' can also be used as provider to explicitly disable tracing.
+  provider: (( grab $TRACING_PROVIDER_NAME || "noop" ))
+
+  # skipTraceExport only applies when provider is stdout. Set skipTraceExport to true
+  # so that trace information is not written to stdout.
+  # skipTraceExport: true
+
+  # endpoint is where trace information should be routed. Applies to zipkin and jaegar.
+  # endpoint: (( grab $TRACING_PROVIDER_ENDPOINT || "http://localhost:9411/api/v2/spans" ))
```
joe94, line 353:
```suggestion
  # endpoint: (( grab $TRACING_PROVIDER_ENDPOINT || "http://zipkin:9411/api/v2/spans" ))
```

caduceus.yaml
```diff
@@ -345,3 +345,17 @@ sender:
...
+  # 'noop' can also be used as provider to explicitly disable tracing.
+  provider: "noop"
+
+  # skipTraceExport only applies when provider is stdout. Set skipTraceExport to true
+  # so that trace information is not written to stdout.
+  # skipTraceExport: true
+
+  # endpoint is where trace information should be routed. Applies to zipkin and jaegar.
+  # endpoint: "http://localhost:9411/api/v2/spans"
+
```
joe94, line 361: A final change is adding this configuration section https://github.com/xmidt-org/tr1d1um/blob/main/tr1d1um.yaml#L234 to 
- `caduceus.yaml`
- `caduceus_spruce.yaml`

deploy/packaging/caduceus_spruce.yaml
```diff
@@ -338,3 +338,18 @@ sender:
...
+
+
```
joe94, line 355:
```suggestion
# timeouts that apply to the Argus HTTP client.
# (Optional) By default, the values below will be used.
argusClientTimeout:
  # clientTimeout is the timeout for requests made through this 
  # HTTP client. This timeout includes connection time, any
  # redirects, and reading the response body.
  clientTimeout: 50s

  # netDialerTimeout is the maximum amount of time the HTTP Client Dialer will
  # wait for a connect to complete.
  netDialerTimeout: 5s
```

caduceus.yaml
joe94, line 361:
```suggestion
# timeouts that apply to the Argus HTTP client.
# (Optional) By default, the values below will be used.
argusClientTimeout:
  # clientTimeout is the timeout for requests made through this 
  # HTTP client. This timeout includes connection time, any
  # redirects, and reading the response body.
  clientTimeout: 50s

  # netDialerTimeout is the maximum amount of time the HTTP Client Dialer will
  # wait for a connect to complete.
  netDialerTimeout: 5s
```

caduceus.yaml
```diff
@@ -345,3 +345,27 @@ sender:
...
+  # 'noop' can also be used as provider to explicitly disable tracing.
+  provider: "noop"
+
+  # skipTraceExport only applies when provider is stdout. Set skipTraceExport to true
+  # so that trace information is not written to stdout.
+  # skipTraceExport: true
+
+  # endpoint is where trace information should be routed. Applies to zipkin and jaegar.
+  # endpoint: "http://localhost:9411/api/v2/spans"
+# timeouts that apply to the Argus HTTP client.
```
joe94, line 361:
```suggestion

# timeouts that apply to the Argus HTTP client.
```

http.go
```diff
@@ -43,9 +43,14 @@ type ServerHandler struct {
 }
 
 func (sh *ServerHandler) ServeHTTP(response http.ResponseWriter, request *http.Request) {
-	debugLog := log.WithPrefix(sh.Logger, level.Key(), level.DebugValue())
-	infoLog := log.WithPrefix(sh.Logger, level.Key(), level.InfoValue())
-	errorLog := log.WithPrefix(sh.Logger, level.Key(), level.ErrorValue())
+	logger := logging.GetLogger(request.Context())
+	if logger == nil {
+	    logger = sh.Logger
+	}
```
kristinapathak, line 49:
```suggestion
```

`logging.GetLogger()` can never return `nil`. Having this check isn't a problem, but it leaves an untestable path in terms of unit test coverage. 😅

main.go
```diff
@@ -52,9 +57,75 @@ var (
...
+type httpClientTimeout struct {
+	// ClientTimeout is HTTP Client Timeout.
+	ClientTimeout time.Duration
+
+
+	// NetDialerTimeout is the net dialer timeout
+	NetDialerTimeout time.Duration
+}
+
+func loadTracing(v *viper.Viper, appName string) (candlelight.Tracing, error) {
```
kristinapathak, line 70: The general organization of caduceus' `main.go` is the `caduceus()` function followed by helper functions for it.  Can you please move the helper functions you've added (`loadTracing()`, `newArgusClientTimeout()`, and `newHTTPClient()`) to be below `func caduceus()`?

main.go
```diff
@@ -132,16 +216,31 @@ func caduceus(arguments []string) int {
 		modifiedWRPCount:         metricsRegistry.NewCounter(ModifiedWRPCounter),
 		maxOutstanding:           0,
 	}
+
 	caduceusConfig.Webhook.Logger = logger
 	caduceusConfig.Webhook.MetricsProvider = metricsRegistry
-	svc, stopWatches, err := ancla.Initialize(caduceusConfig.Webhook, caduceusSenderWrapper)
+	argusClientTimeout, err := newArgusClientTimeout(v)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Unable to parse argus client timeout config values: %s \n", err.Error())
```
kristinapathak, line 224:
```suggestion
		fmt.Fprintf(os.Stderr, "Unable to parse argus client timeout config values: %v\n", err)
```

Suggested extremely minor change to match the other `Fprintf()` statements.

main.go
```diff
@@ -63,6 +134,7 @@ func caduceus(arguments []string) int {
 		v = viper.New()
 
 		logger, metricsRegistry, webPA, err = server.Initialize(applicationName, arguments, f, v, Metrics, ancla.Metrics)
+		infoLogger                          = level.Info(logger)
```
kristinapathak, line 137: I don't have a problem with adding `infoLogger`, but there are plenty of places in `caduceus()` where we are still logging with `logger.Log(level.Key(), level.InfoValue(),...)`.  I think we should be consistent - so either log all info level logs using `infoLogger` or don't create a specific `infoLogger`.

main.go
```diff
@@ -132,16 +216,31 @@ func caduceus(arguments []string) int {
...
 		return 1
 	}
+	infoLogger.Log(logging.MessageKey(), "Webhook service enabled")
+
+	rootRouter := mux.NewRouter()
+	otelMuxOptions := []otelmux.Option{
+		otelmux.WithPropagators(tracing.Propagator),
+		otelmux.WithTracerProvider(tracing.TracerProvider),
+	}
+	rootRouter.Use(otelmux.Middleware("mainSpan", otelMuxOptions...), candlelight.EchoFirstTraceNodeInfo(tracing.Propagator))
```
kristinapathak, line 241: @joe94 Is `mainSpan` the name we want for this server? I'm not sure the repercussions of this, but I would think either `caduceus` or `primary`?

kristinapathak: Or maybe take the value from `candelight.Config.ApplicationName`?
https://pkg.go.dev/github.com/xmidt-org/candlelight#Config

kristinapathak (changes requested):
Overall code flow looks good; gave some suggestions that I think would make this even better. 👍

main.go
utsavbatra5: @joe94 can you please clarify on this?

main.go
utsavbatra5: I think we have released all the other microservices with the same name. So we can keep it like this for now.

http.go
utsavbatra5: I have changed it to logging.DefaultLogger() instead of nil in the latest commit.

main.go
utsavbatra5: Taken care of in latest commit.

main.go
utsavbatra5: Taken care of in latest commit.

http.go
Sachin4403: @kristinapathak  now we are comparing it is the default logger instead of nil, we haven't updated  if else block because we don't want to use the default logger which is provided by logging package.

Here also we did the same thing https://github.com/xmidt-org/argus/pull/145
Please refer to last conversation

main.go
joe94: I agree with @kristinapathak about staying consistent. 

I would suggest updating this method so leveled log calls are of the form `level.[LEVEL](logger).Log(...)` (i.e. instead of using a variable, just do something like `level.Info(logger).Log(logging.MessageKey(), "Webhook service enabled")`). Like suggested in the docs: https://github.com/go-kit/kit/blob/master/log/level/doc.go#L14

Other examples to udpate:
1) `log.WithPrefix(logger, level.Key(), level.InfoValue()).Log("configurationFile", v.ConfigFileUsed())` **should be** 
`level.Info(logger).Log("configurationFile", v.ConfigFileUsed())`

2) `logger.Log(level.Key(), level.ErrorValue(), logging.MessageKey(), "Unable to initialize service discovery environment", logging.ErrorKey(), err)` **should be** 
`level.Error(logger).Log(logging.MessageKey(), "Unable to initialize service discovery environment", logging.ErrorKey(), err)`

main.go
joe94: Yeah, I called it `mainSpan` in the earlier services we integrated because we were only instrumenting one server. I just saw, though, that in Talaria we have both the `primary` and the `control` servers so I think it'll be good to use the server names from now on. 

@utsavbatra5, let's rename this to `primary`. We can retroactively update the other services as we go. 

From what I saw using the Zipkin UI, the node name (which comes from `candlelight.Config.ApplicationName`) is the most important visual cue in the trace graph. This operation name will mostly help when inspecting specific requests.

main.go
utsavbatra5: I have fixed the same in latest commit.

main.go
utsavbatra5: I have fixed the same in latest commit.

joe94 (approved):
LGTM!

kristinapathak (approved):
Looks great! 🚀

PR #264, opened 2021-04-30
Add the Github license checker action to the CI pipeline

schmidtw:
Add the Github license checker action to the CI pipeline.  Also I went through the licenses and approved the ones where the text wasn't entirely identified.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -107,6 +107,7 @@ jobs:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
 
   licensed:
+    name: License Checker
     runs-on: [ ubuntu-latest ]
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -21,9 +21,17 @@ jobs:
         with:
           go-version: '1.15.x' # The Go version to download (if necessary) and use.
 
+      # Setup the license checker
+      - uses: jonabc/setup-licensed@v1
+        with:
+          version: 2.x
+
       # Run build of the application
       - name: Run build
         run: go build ./...
+
+      - name: Run License Check
+        run: licensed status
 
       # Run gofmt on the code
       - name: Run gofmt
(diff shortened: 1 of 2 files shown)


schmidtw:
Please squash & merge this when merging.

.github/workflows/push.yml
```diff
@@ -105,3 +105,19 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+
+  licensed:
+    name: License Checker
+    runs-on: [ ubuntu-latest ]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
```
kristinapathak, line 114: Do we want to lock into a go version (like in the test job above)?

.github/workflows/push.yml
```diff
@@ -105,3 +105,19 @@ jobs:
...
+      - uses: jonabc/setup-licensed@v1
+        with:
+         version: 3.x
+
+      - run: go get
+
```
kristinapathak, line 120:
```suggestion
      - run: go get
```

Do we need these spaces?

.licensed.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - mit 
+  - apache-2.0
+  - bsd-2-clause
+  - bsd-3-clause
+  - isc
+  - mpl-2.0         # Needed for hashicorp stuff
+
+
+reviewed:
+  go:
```
kristinapathak, line 11: Do we need these listed here? From the documentation [here](https://github.com/github/licensed/blob/master/docs/configuration.md#applications), it sounds like this area is for dependencies that have licenses not in the allowed list, but all of these don't follow that guidance?

schmidtw: I'll try removing them & see if v3 does a better job resolving the licenses.  These all showed up as `other` & failed the build.

kristinapathak: Oh, strange...

kristinapathak (approved):
looks good! Once we merge this, we can work on adding it to all other repos and our template.

PR #271, opened 2021-07-09
updated webpa-common, argus, ancla versions

kristinapathak:


main.go
```diff
@@ -165,7 +165,7 @@ func caduceus(arguments []string) int {
 	}
 
 	caduceusConfig.Webhook.Argus.HTTPClient = newHTTPClient(argusClientTimeout, tracing)
-	svc, stopWatches, err := ancla.Initialize(caduceusConfig.Webhook, getLogger, caduceusSenderWrapper)
+	svc, stopWatches, err := ancla.Initialize(caduceusConfig.Webhook, getLogger, logging.WithLogger, caduceusSenderWrapper)
```
joe94, line 168: are we assuming here that getLogger must be compatible with `logging.WithLogger`?

kristinapathak: Yes, it's the caller's responsibility to make sure ancla will work.  Before, we were just assuming `getLogger` would work with the way ancla was setting the logger, with the logger not having any visibility into how it was being set.  Both have their own sets of problems... :shrug:

joe94: yeah, that sounds fair to me 👍🏼

joe94 (approved):

Issue #272, opened 2021-08-09
Submitting event with header returns 404


ilawjr:
If you attempt to submit a valid event to the notify endpoint without this header:
`Content-type:application/msgpack`
You will receive a 404 ( Page Not Found or File Not Found) instead of a 400 (Bad Request).

kristinapathak:
This is the place in the code where the mux router is given the header logic:
https://github.com/xmidt-org/caduceus/blob/d011ae1a4af4aee68e7cecc8968a7ad2a40a0948/primaryHandler.go#L57

Because not including that header means the mux router doesn't match to a handler, the 404 is returned.  If we add the logic into the handler itself, then we can return a 400 like we want:
https://github.com/xmidt-org/caduceus/blob/d011ae1a4af4aee68e7cecc8968a7ad2a40a0948/http.go#L45

There may be a way to do this with the mux router and I just don't know how.

schmidtw:
Should this be a `415` (`Unsupported Media Type`) vs. a more generic `400`?

PR #273, opened 2021-08-12
Moved content-type header check

mtrinh11:
Added Content-type header check in http.go


mtrinh11:
fixes https://github.com/xmidt-org/caduceus/issues/272

kristinapathak (approved):

PR #276, opened 2021-09-24
added initial checks for partnerIDs

mtrinh11:


outboundSender.go
```diff
@@ -368,6 +387,28 @@ func (obs *CaduceusOutboundSender) Queue(msg *wrp.Message) {
 		return
 	}
 
+	//check the partnerIDs
+	if len(obs.listener.PartnerIDs) == 0 {
+		switch obs.noPIDAction {
```
johnabass, line 392: Consider making `noPIDAction` an enumerated type, and move the `switch/case` into a method of that type.  Here's one approach:

```
type NoPIDAction string

const (
    Custom NoPIDAction = "custom"
    DoNotSend NoPIDAction = "donotsend"
)

func (npa NoPIDAction) Execute(l Listener, pids []string) {
    switch {
        case Custom:
            l.PartnerIDs = pids
        case DoNotSend:
            // you could omit this "do nothing" case and just have a default that does nothing
        default:
            // can panic, return an error, do nothing, etc
    }
}
```

The advantage to this is that all of the logic for deciding what to do for a particular action is encapsulated in one place.  That reduces code noise and footprint of calling code, and it makes things easier to unit test.

outboundSender.go
```diff
@@ -368,6 +387,28 @@ func (obs *CaduceusOutboundSender) Queue(msg *wrp.Message) {
...
+		case "custom":
+			obs.listener.PartnerIDs = obs.customPIDs
+		case "donotsend":
+			return
+		}
+
+	} else {
+		hasPID := false
```
johnabass, line 400: Consider moving this logic to a helper function, probably an unexported function.  That makes this logic easier to unit test, and reduces the code noise of the enclosing function.

Alternatively, you could make this a method of whatever `obs.listener` is.

caduceus.yaml
```diff
@@ -351,6 +351,17 @@ sender:
 #   - 1000
 #   - 10000
 
+  # noPIDAction dictates what to do if PartnerIDs are missing or empty. "donotsend"
```
kristinapathak, line 354: Similar to tr1d1um, we want a boolean that tells us whether or not to enforce the partner ID check (with `false` being enforcing).

caduceus.yaml
```diff
@@ -351,6 +351,17 @@ sender:
...
+  # will have Caduceus not send the msg to all listeners if PartnerIDs are not
```
kristinapathak, line 355:
```suggestion
  # will have Caduceus not send the msg to any listeners if PartnerIDs are not 
```

outboundSender.go
```diff
@@ -351,6 +361,15 @@ func (obs *CaduceusOutboundSender) RetiredSince() time.Time {
 	return deliverUntil
 }
 
+func contains(str string, sl []string) bool {
```
kristinapathak, line 364: Instead of taking a `string`, and `[]string`, you can make a function `overlaps()` that takes two `[]string`s and returns `true` when there is at least one shared string value.

kristinapathak (commented):
Looks good so far! Tried not to overlap with John's comments 🙂

outboundSender.go
mtrinh11: after today's conversation and some clarification from Kristina, we're going to get rid of this switch and simplify to just 2 use cases.

caduceus.yaml
```diff
@@ -351,6 +351,14 @@ sender:
 #   - 1000
 #   - 10000
 
+  # customPIDs is a custom list ofallowed PartnerIDs that will be used if noPIDAction 
+  # is "custom".
```
kristinapathak, line 355:
```suggestion
  # customPIDs is a custom list of allowed PartnerIDs that will be used if a message has no partner IDs.  When empty, a message with no partner IDs will not be sent to any listeners when enforcing the partner ID check.
```

caduceus.yaml
```diff
@@ -351,6 +351,14 @@ sender:
...
+  customPIDs: []
+
+  # disablePartnerIDs dictates whether or not to enfore the partnerID check
```
kristinapathak, line 358:
```suggestion
  # disablePartnerIDs dictates whether or not to enforce the partnerID check.
```

senderWrapper.go
```diff
@@ -65,10 +65,16 @@ type SenderWrapperFactory struct {
 
 	// The http client Do() function to share with OutboundSenders.
 	Sender func(*http.Request) (*http.Response, error)
+
+	// A custom list of accepted PartnerIDs that are used if NoPIDAction is of value "custom".
```
kristinapathak, line 69: Please update this comment.

senderWrapper.go
```diff
@@ -65,10 +65,16 @@ type SenderWrapperFactory struct {
...
+	CustomPIDs []string
+
+	// Dictates whether or not to enforce the partner ID check.
```
kristinapathak, line 72:
```suggestion
	// DisablePartnerIDs dictates whether or not to enforce the partner ID check.
```

outboundSender.go
```diff
@@ -99,10 +99,16 @@ type OutboundSenderFactory struct {
 
 	// The logger to use.
 	Logger log.Logger
+
+	// A custom list of accepted PartnerIDs that are used if NoPIDAction is of value "custom".
```
kristinapathak, line 103: This comment should also be updated.

outboundSender.go
```diff
@@ -99,10 +99,16 @@ type OutboundSenderFactory struct {
...
+	CustomPIDs []string
+
+	// Dictates whether or not to enforce the partner ID check.
```
kristinapathak, line 106:
```suggestion
	// DisablePartnerIDs dictates whether or not to enforce the partner ID check.
```

kristinapathak (approved):
Looks good! Just a few comment/documentation suggestions.

johnabass (approved):
👍

PR #280, opened 2021-10-28
changed retry funcs

mtrinh11:


outboundSender.go
```diff
@@ -622,15 +622,8 @@ func (obs *CaduceusOutboundSender) send(urls *ring.Ring, secret, acceptType stri
...
-		ShouldRetryStatus: func(code int) bool {
-			for _, c := range obs.retryCodes {
-				if code == c {
-					return true
-				}
-			}
-			return false
-		},
+		ShouldRetry:       xhttp.ShouldRetry,
+		ShouldRetryStatus: xhttp.ShouldRetryStatus,
```
Sachin4403, line 626: @mtrinh11  the function name should be **RetryCodes** as the same is defined in  https://github.com/xmidt-org/webpa-common/blob/a59029ee5adce57ab9671e5efc0c817f3675d8a4/xhttp/retry.go#L187

kristinapathak (approved):

PR #312, opened 2022-05-04
Track latency

### LINKED ISSUE #100: We need to track latency for calls to the outbound client URLs
Right now we can't track the time spent waiting on outbound client URLs to respond.  This causes the problem where network issues in between the client and caduceus are difficult to diagnose.

Ideally we will implement something like this to track the time in a useful way.

https://godoc.org/github.com/prometheus/client_golang/prometheus/promhttp#example-InstrumentRoundTripperDuration

renaz6:
- Tracks the time spent waiting on outbound client URLs to respond (latency)
- Creates a new histogram metric for Caduceus called `query_duration_seconds_histogram`
- closes #100


HttpClient.go
```diff
@@ -0,0 +1,60 @@
...
+		startTime := m.now()
+		resp, err := next.Do(req)
+		endTime := m.now()
+		var code string
+
+		if err != nil {
+			code = networkError
+		}
+
+		code = strconv.Itoa(resp.StatusCode)
```
kristinapathak, line 52:
```suggestion
		code := networkError

		if err == nil {
			code = strconv.Itoa(resp.StatusCode)
		}
```

outboundSender.go
```diff
@@ -656,7 +673,11 @@ func (obs *CaduceusOutboundSender) send(urls *ring.Ring, secret, acceptType stri
 		"event.source", msg.Source,
 		"event.destination", msg.Destination,
 	)
-	resp, err := xhttp.RetryTransactor(retryOptions, obs.sender)(req)
+
+	retryer := xhttp.RetryTransactor(retryOptions, obs.sender.Do)
+	resp, err := retryer(req)
```
kristinapathak, line 678:
```suggestion
	retryer := xhttp.RetryTransactor(retryOptions, obs.sender.Do)
	client := obs.clientMiddleware(retryer)
	resp, err := client.Do(req)
```

senderWrapper.go
```diff
@@ -61,14 +60,16 @@ type SenderWrapperFactory struct {
...
+	Sender httpClient
 
 	// CustomPIDs is a custom list of allowed PartnerIDs that will be used if a message
 	// has no partner IDs.
 	CustomPIDs []string
 
 	// DisablePartnerIDs dictates whether or not to enforce the partner ID check.
 	DisablePartnerIDs bool
+
+	Querylatency metrics.Histogram
```
kristinapathak, line 72:
```suggestion
```
The senderWrapper should make the metric using the MetricsRegistry.

senderWrapper.go
```diff
@@ -91,10 +92,12 @@ type CaduceusSenderWrapper struct {
 	senders             map[string]OutboundSender
 	metricsRegistry     CaduceusMetricsRegistry
 	eventType           metrics.Counter
+	queryLatency        metrics.Histogram
 	wg                  sync.WaitGroup
 	shutdown            chan struct{}
 	customPIDs          []string
 	disablePartnerIDs   bool
+	clientMiddleware    func(httpClient) httpClient
```
kristinapathak, line 100:
```suggestion
```

We can't store this because it will be new for each outbound sender.

senderWrapper.go
```diff
@@ -120,6 +123,9 @@ func (swf SenderWrapperFactory) New() (sw SenderWrapper, err error) {
 		return
 	}
 
+	measure := NewMetricWrapperMeasures(swf.MetricsRegistry)
```
kristinapathak, line 126:
```suggestion
```

mocks_test.go
```diff
@@ -95,6 +95,25 @@ func (m *mockGauge) With(labelValues ...string) metrics.Gauge {
...
+}
+
+func (m *mockHistogram) With(labelValues ...string) metrics.Histogram {
+	for _, v := range labelValues {
+		if !utf8.ValidString(v) {
+			panic("not UTF-8")
+		}
+	}
+	args := m.Called(labelValues)
+	return args.Get(0).(metrics.Histogram)
```
kristinapathak, line 114:
```suggestion
	return m
```
I don't think we'll want something special returned here, so the easiest thing to do is return our same mock - then the same mock can expect `With()` and `Observe()` calls.

HttpClient.go
```diff
@@ -0,0 +1,58 @@
+package main
```
kristinapathak, line 1: Can you rename this file to `httpClient.go`?

HttpClient.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/go-kit/kit/metrics"
+)
+
+type httpClient interface {
+	Do(*http.Request) (*http.Response, error)
+}
+
+// DoerFunc implements HTTPClient
+type DoerFunc func(*http.Request) (*http.Response, error)
```
kristinapathak, line 17:
```suggestion
type doerFunc func(*http.Request) (*http.Response, error)
```

I don't think we need to export this type.

main.go
```diff
@@ -132,10 +132,10 @@ func caduceus(arguments []string) int {
 		DeliveryInterval:    caduceusConfig.Sender.DeliveryInterval,
 		MetricsRegistry:     metricsRegistry,
 		Logger:              logger,
-		Sender: (&http.Client{
+		Sender: DoerFunc((&http.Client{
```
kristinapathak, line 135:
```suggestion
		Sender: DoerFunc(&http.Client{
```
Do we need this additional parenthesis?

outboundSender.go
```diff
@@ -103,6 +107,8 @@ type OutboundSenderFactory struct {
 
 	// DisablePartnerIDs dictates whether or not to enforce the partner ID check.
 	DisablePartnerIDs bool
+
+	QueryLatency metrics.Histogram
```
kristinapathak, line 111: Do we need this field?

outboundSender.go
```diff
@@ -142,6 +148,7 @@ type CaduceusOutboundSender struct {
 	maxWorkersGauge                  metrics.Gauge
 	currentWorkersGauge              metrics.Gauge
 	deliveryRetryMaxGauge            metrics.Gauge
+	querylatency                     metrics.Histogram
```
kristinapathak, line 151: Do we need this field?

outboundSender.go
```diff
@@ -225,6 +238,10 @@ func (osf OutboundSenderFactory) New() (obs OutboundSender, err error) {
 	return
 }
 
+func NopHTTPClient(next httpClient) httpClient {
+	return next
+}
```
kristinapathak, line 243: Can you move this to httpClient.go?

senderWrapper.go
```diff
@@ -132,6 +133,10 @@ func (swf SenderWrapperFactory) New() (sw SenderWrapper, err error) {
 	return
 }
 
+func NewMetricWrapperMeasures(m CaduceusMetricsRegistry) metrics.Histogram {
+	return m.NewHistogram(QueryDurationSecondsHistogram, 11)
+}
```
kristinapathak, line 138: Can you move this to metrics.go?

kristinapathak (commented):
Looks good, mostly suggesting minor tweaks and moving code around.  The httpClient needs unit tests, which I know you're still working on. 🙂

main.go
renaz6: I think the parenthesis is needed. The call breaks without it. The inner set of parentheses ensures that Do is called on the http.Client

outboundSender.go
renaz6: I don't think so. It was unused. Thanks!

outboundSender.go
renaz6: I think it's needed. This field is used in senderWrapper.go to add the "url" label to the histogram

httpClient_test.go
```diff
@@ -0,0 +1,86 @@
...
+			}
+
+			c := m.roundTripper(client)
+			resp, err := c.Do(tc.request)
+
+			// Check response
+			assert.Equal(t, resp.StatusCode, tc.expectedResponse)
+
+			// Check Error
+			assert.Equal(t, err, nil)
```
kristinapathak, line 78: We'll need to update this when an error is possible - when we expect no errors, we can use `assert.NoError()` and when we expect a specific error we can use `assert.ErrorsIs()`. 👍

kristinapathak (commented):
Good start on the tests, left some suggestions 👍

httpClient_test.go
```diff
@@ -75,7 +88,11 @@ func TestRoundTripper(t *testing.T) {
...
 			// Check Error
-			assert.Equal(t, err, nil)
+			if tc.expectedErr != nil {
+				assert.ErrorIs(t, tc.expectedErr, err)
+			} else {
+				assert.NoError(t, err)
+			}
 
 			// Check the histogram
 			fakeHandler.AssertExpectations(t)
```
kristinapathak, line 98:
```suggestion
			fakeHandler.AssertExpectations(t)
			fakeHistogram.AssertExpectations(t)
```

httpClient_test.go
```diff
@@ -34,39 +37,49 @@ func TestRoundTripper(t *testing.T) {
...
-		fakeHist := new(mockHistogram)
-		fakeHist.On("With", mock.AnythingOfType("[]string")).Return().Times(2)
-		fakeHist.On("Observe", mock.AnythingOfType("float64")).Return().Times(2)
-
 		t.Run(tc.description, func(t *testing.T) {
 
+			fakeTime := mockTime(tc.startTime, tc.endTime)
+			fakeHandler := new(mockHandler)
+			fakeHist := new(mockHistogram)
+			fakeHist.On("With", mock.AnythingOfType("[]string")).Return().Times(2)
```
kristinapathak, line 67: Does this work?
```go
			fakeHist.On("With", "code", tc.expectedResponse).Return().Once()
			fakeHist.On("With", "url", ?????).Return().Once() // I don't think this With is being called? I can't figure out what the url might be.
```

httpClient_test.go
```diff
@@ -34,39 +37,49 @@ func TestRoundTripper(t *testing.T) {
...
+			fakeHist.On("Observe", mock.AnythingOfType("float64")).Return().Times(2)
```
kristinapathak, line 68: Do we really call observe twice?

kristinapathak (commented):
This is looking great.  A few minor tweaks and I think it's good to go!

kristinapathak (approved):
Awesome! 😄

PR #314, opened 2022-05-17
update senderwrapper to defer mutex unlocking

kristinapathak:


senderWrapper.go
```diff
@@ -238,3 +230,21 @@ func undertaker(sw *CaduceusSenderWrapper) {
...
 }
+
+func createDeadlist(sw *CaduceusSenderWrapper, threshold time.Time) map[string]OutboundSender {
+	if sw == nil || threshold.IsZero() {
+		return nil
+	}
+
+	deadList := make(map[string]OutboundSender)
+	sw.mutex.Lock()
+	defer sw.mutex.RUnlock()
```
johnabass, line 241: This is mismatched.  `Lock` needs `Unlock`, and `RLock` needs `RUnlock`.  As written here, the `defer` will possibly panic since there was no corresponding acquisition of the write lock.  But, for sure this will deadlock since there is no code releasing the write lock acquired via `Lock`.

johnabass (changes requested):

johnabass (approved):

johnabass (approved):

PR #323, opened 2022-06-08
Add Webhook URI in latency metric of the outbound sender

chaitanyasingla-dt:
There has been an addition in caduceus to track latency metrics for outbound senders.

Scenario:
Let's suppose we have multiple webhooks getting tracked and only one of them is increasing the latency of the request. If all the webhooks latency metric is getting tracked under the same name that is ```code = strconv.Itoa(resp.StatusCode)```, we will not be able to identify the source of the increase in latency of the requests.

Solution:
This pull request adds Webhook URI to the latency metric to identify the source of the increase in the latency of the requests.


denopink:
work completed in #461 , the duration metric will now have a `url` label

PR #324, opened 2022-06-14
Incoming queue latency

### LINKED ISSUE #319: Track latency metric for time spent in "incoming queue"
- We need to add a new latency metric to track time spent in the "incoming queue" (ie how long it took to decode and then try to queue in all outbound senders)

renaz6:
- Added new latency metric to track time spent in the "incoming queue" (ie how long it took to decode and then try to queue in all outbound senders)
- Closes #319

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,10 @@ _testmain.go
 # IDE files
 .idea/*
 
+# VSCode
+*.code-workspace
+.vscode/*
+
 caduceus
 .ignore
 
--- a/http.go
+++ b/http.go
@@ -42,11 +42,11 @@ type ServerHandler struct {
 	incomingQueueDepth       int64
 	maxOutstanding           int64
 	incomingQueueLatency     metrics.Histogram
-	startTime                time.Time
+	now                      func() time.Time
 }
 
 func (sh *ServerHandler) ServeHTTP(response http.ResponseWriter, request *http.Request) {
-	sh.startTime = time.Now()
+	startTime := sh.now()
 	eventType := "Unknown"
 	logger := logging.GetLogger(request.Context())
 	if logger == logging.DefaultLogger() {
@@ -61,15 +61,14 @@ func (sh *ServerHandler) ServeHTTP(response http.ResponseWriter, request *http.R
 
 	infoLog.Log(messageKey, "Receiving incoming request...")
 
+	// find time difference, add to metric after function finishes
+	defer addToHistogram(sh, eventType, startTime)
+
 	if len(request.Header["Content-Type"]) != 1 || request.Header["Content-Type"][0] != "application/msgpack" {
 		//return a 415
 		response.WriteHeader(http.StatusUnsupportedMediaType)
 		response.Write([]byte("Invalid Content-Type header(s). Expected application/msgpack. \n"))
 		debugLog.Log(messageKey, "Invalid Content-Type header(s). Expected application/msgpack. \n")
-
-		// find time difference, add to metric
-		latency := time.Since(sh.startTime)
-		sh.incomingQueueLatency.With("event type", eventType).Observe(latency.Seconds())
 		return
 	}
 
@@ -81,10 +80,6 @@ func (sh *ServerHandler) ServeHTTP(response http.ResponseWriter, request *http.R
 		response.WriteHeader(http.StatusServiceUnavailable)
 		response.Write([]byte("Incoming queue is full.\n"))
 		debugLog.Log(messageKey, "Incoming queue is full.\n")
-
-		// find time difference, add to metric
-		latency := time.Since(sh.startTime)
-		sh.incomingQueueLatency.With("event type", eventType).Observe(latency.Seconds())
 		return
 	}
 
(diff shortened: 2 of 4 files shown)


http.go
```diff
@@ -40,9 +41,13 @@ type ServerHandler struct {
 	modifiedWRPCount         metrics.Counter
 	incomingQueueDepth       int64
 	maxOutstanding           int64
+	incomingQueueLatency     metrics.Histogram
+	now                      func() time.Time
 }
 
 func (sh *ServerHandler) ServeHTTP(response http.ResponseWriter, request *http.Request) {
+	startTime := sh.now()
+	eventType := "Unknown"
```
kristinapathak, line 50: I would suggest making this an unexported const at the top of `metrics.go`.  Then the tests can use it too.

```go
     unknownEventType = "unknown"
```

.gitignore
```diff
@@ -45,6 +45,10 @@ _testmain.go
 # IDE files
 .idea/*
 
+# VSCode
+*.code-workspace
+.vscode/*
+
```
denopink, line 51: Don't mind me, I'm a VSCode user 👻

http_test.go
```diff
@@ -70,23 +70,34 @@ func exampleRequest(msgType int, list ...string) *http.Request {
 
 func TestServerHandler(t *testing.T) {
 	fmt.Print("TestingeServerHandler")
```
denopink, line 72: Do we care if remove these old random print statements? There's one per test

metrics.go
```diff
@@ -147,6 +149,13 @@ func Metrics() []xmetrics.Metric {
 			LabelNames: []string{"url", "code"},
 			Buckets:    []float64{0.0625, 0.125, .25, .5, 1, 5, 10, 20, 40, 80, 160},
 		},
+		{
+			Name:       IncomingQueueLatencyHistogram,
+			Help:       "A histogram of latencies for the incoming queue.",
+			Type:       "histogram",
+			LabelNames: []string{"code"},
+			Buckets:    []float64{0.0625, 0.125, .25, .5, 1, 5, 10, 20, 40, 80, 160},
```
denopink, line 157: Should add the latency unit in the help description?

denopink (approved):
lgtm! 🍻 

I just had some super minor comments, but nothing that'll affect the PR.

mtrinh11 (approved):
lgtm!

http.go
```diff
@@ -56,6 +61,12 @@ func (sh *ServerHandler) ServeHTTP(response http.ResponseWriter, request *http.R
 
 	infoLog.Log(messageKey, "Receiving incoming request...")
 
+	// find time difference, add to metric after function finishes
+	defer func() {
```
johnabass, line 65: `defer` closures that are inline can get a bit hard to read after a while, especially as they are added over time.

A cleaner way is just pull out the `defer` into an unexported method.  This has the extra side effect of making the defer logic separately testable.

For example:

```
// this function can now have separate unit tests, if applicable
// especially as the codebase changes over time, this separate function is easier to deal with
func (sh *ServerHandler) recordQueueLatency(startTime time.Time) {
    endTime := sh.now()
    sh.incomingQueueLatency.With("event", eventType).Observe(endTime.Sub(startTime).Seconds())
}

func (sh *ServerHandler) ServeHTTP(response http.ResponseWriter, request *http.Request) {
    startTime := sh.now()
    defer sh.recordQueueLatency(startTime)

    // if you don't need the startTime variable below, you can just abbreviate the above to:
    defer sh.recordQueueLatency(sh.now())

    // the rest of the code as is
}
```

johnabass (changes requested):

denopink (approved):
lgtm 🍻

johnabass (approved):

metrics.go
```diff
@@ -6,32 +6,34 @@ import (
...
+	DropsDueToInvalidPayload              = "drops_due_to_invalid_payload"
+	OutgoingQueueDepth                    = "outgoing_queue_depths"
+	DropsDueToPanic                       = "drops_due_to_panic"
+	ConsumerRenewalTimeGauge              = "consumer_renewal_time"
+	ConsumerDeliverUntilGauge             = "consumer_deliver_until"
+	ConsumerDropUntilGauge                = "consumer_drop_until"
+	ConsumerDeliveryWorkersGauge          = "consumer_delivery_workers"
+	ConsumerMaxDeliveryWorkersGauge       = "consumer_delivery_workers_max"
+	QueryDurationSecondsHistogram         = "query_duration_seconds_histogram"
+	IncomingQueueLatencyHistogram_seconds = "incoming_queue_latency_histogram"
```
denopink, line 28: did we mean to put `_seconds` in the actual value of `IncomingQueueLatencyHistogram_seconds`?

http.go
```diff
@@ -40,9 +41,17 @@ type ServerHandler struct {
...
+	incomingQueueLatency     metrics.Histogram
+	now                      func() time.Time
 }
 
 func (sh *ServerHandler) ServeHTTP(response http.ResponseWriter, request *http.Request) {
+	eventType := unknownEventType
+	// find time difference, add to metric after function finishes
+	defer func() {
+		sh.recordQueueLatencyToHistogram(sh.now(), eventType)
+	}()
```
denopink, line 53: don't we want to execute `sh.now()` outside of the defer, otherwise it'll be execute at the end of `ServeHTTP` and not at the beginning?
like so
```go
defer func(s) {
	sh.recordQueueLatencyToHistogram(s, eventType)
}(sh.now())
```

PR #428, opened 2023-11-02
added a metrics package and moved metrics files to package

maurafortino:
What's Included:

- created a metrics package
  - moved metrics.go, metrics_test.go and anclaHelper.go files to the metrics package
  - moved CaduceusMetricsRegistry interface to metrics.go file (import issues if left in main)
  - moved CreateOutbounderMetrics to outboundSender.go  (import issues if left in metrics package)
- moved error reasons constants to http.go (all but one are only related to http.go - only error related to metrics is network error) 
- all other updates are just updating files with the metrics package and differentiating between caduceus metrics package and gokit metrics package.


denopink (approved):
I made a quick change. otherwise, lgtm!

PR #436, opened 2024-01-16
Senderwrapper

maurafortino:
What's Included:
- commented back in caduceus request handler in http.go to implement the handler
- updated senderwrapper to use uber/fx
- created fx module for senderwrapper
- removed modules and replaced them with provide functions based off of uber/fx best practices
- Kept OutboundSenderFactory to avoid passing mutexes, but simplified so there wasn't so much duplication from SenderConfig --> CaduceusSenderWrapper --> OutboundSenderFactory --> CadueceusOutboundSender
- updated functions for metrics of senderwrapper and outboundsender
- added a webhook stub for testing


denopink (approved):
lgtm but I recommend (if we aren't already doing this) that we reconsider not using the preexisting design (e.g. CaduceusSenderWrapper and etc) and try simplifying the overall outbound sender design and the related support structs

go.mod
```diff
@@ -8,7 +8,6 @@ require (
 	emperror.dev/emperror v0.33.0
 	github.com/alecthomas/kong v0.8.1
 	github.com/go-chi/chi/v5 v5.0.10
-	github.com/go-kit/kit v0.13.0
```
denopink, line 11: nice!

PR #451, opened 2024-02-15
Feat/senders v2

maurafortino:
What's Included:
- added new schema structs (ListenerStub generic in another PR)
- updated wrapper and sender name to reflect new functionality; removed outboundfactory (no longer needed)
- moved listener out of the wrapper struct (still considering where the client struct/client middleware will best be initiated so remaining in wrapper struct for now) 
- added some comments


routes.go
```diff
@@ -68,18 +68,17 @@ func provideCoreOption(server string, in RoutesIn) arrangehttp.Option[http.Serve
...
-			mux.Use(recovery.Middleware(recovery.WithStatusCode(555)))
-
 			options := []otelmux.Option{
 				otelmux.WithTracerProvider(in.Tracing.TracerProvider()),
 				otelmux.WithPropagators(in.Tracing.Propagator()),
 			}
 
+			// TODO: should probably customize things a bit
+			mux.Use(recovery.Middleware(recovery.WithStatusCode(555)), otelmux.Middleware("server_primary", options...),
+				candlelight.EchoFirstTraceNodeInfo(in.Tracing.Propagator(), true))
```
denopink, line 78: this will cause an annoying bug
```suggestion
				candlelight.EchoFirstTraceNodeInfo(in.Tracing.Propagator(), false))
```

sinkSender.go
```diff
@@ -40,59 +43,51 @@ const failureText = `Unfortunately, your endpoint is not able to keep up with th
...
-	Text string `json:"text"`
-	// Original     ancla.InternalWebhook `json:"webhook_registration"` //TODO: add back in once ancla/argus dependency issue is fixed.
-	CutOffPeriod string `json:"cut_off_period"`
-	QueueSize    int    `json:"queue_size"`
-	Workers      int    `json:"worker_count"`
+	Text         string       `json:"text"`
+	Original     ListenerStub `json:"webhook_registration"` //TODO: remove listener stub once ancla/argus issues fixed
+	CutOffPeriod string       `json:"cut_off_period"`
+	QueueSize    int          `json:"queue_size"`
+	Workers      int          `json:"worker_count"`
```
denopink, line 50: we should consider if keeping the original struct makes sense... like if `Original     ListenerStub` is not used then we should get rid of it

just a general suggestion

denopink (approved):
left a few comments, nothing big since we're still working on the core code on later prs

PR #456, opened 2024-02-22
Listener interface draft solution

maurafortino:
This is most likely not the final solution.

Need to workshop


denopink:
closing since we're not going with this idea

PR #476, opened 2024-04-08
Feat/sink interface

maurafortino:
1.  sinkwrapper client interface will be replace by the  sink interface
2. the sink interface will have the following funcs (`update(l Listener)`, `send(urls *ring.Ring, secret, acceptType string, msg *wrp.Message)`)
3. the webhook abstraction will be renamed to something like matcher since it's main purpose is to match incoming messages to desired events or wrp field regex

### DIFF
--- a/sink.go
+++ b/sink.go
@@ -1,0 +1,112 @@
+package main
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/xmidt-org/wrp-go/v3"
+	"go.uber.org/zap"
+)
+
+type SinkI interface {
+	Update(Listener) error
+	Shutdown(bool)
+	Queue(*wrp.Message)
+}
+
+type Sink struct {
+}
+
+func (s *Sink) Update(l Listener) (err error) {
+	switch v := l.(type) {
+	case *ListenerV1:
+		m := &MatcherV1{}
+		if err = m.Update(*v); err != nil {
+			return
+		}
+		s.matcher = m
+	default:
+		err = fmt.Errorf("invalid listner")
+	}
+
+	s.renewalTimeGauge.Set(float64(time.Now().Unix()))
+
+	s.mutex.Lock()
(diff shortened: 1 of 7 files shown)


sink.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/xmidt-org/wrp-go/v3/wrphttp"
+	"go.uber.org/zap"
+)
+
+type SinkI interface {
+	Update(Listener) error
+	Send(*ring.Ring, string, string, *wrp.Message) error
+}
+
+type SinkWebhookV1 struct {
```
denopink, line 29:
```suggestion
type WebhookV1 struct {
```
later on if you move this file to it's own package, e.g. `sink` package, this name might stutter. e.g. `sink.SinkWebhookV1` vs `sink.WebhookV1`

sinkSender.go
```diff
@@ -54,17 +48,31 @@ type Sender interface {
...
+	dropUntil         time.Time
+	deliveryInterval  time.Duration
+	cutOffPeriod      time.Duration
+	queue             atomic.Value
+	wg                sync.WaitGroup
+	workers           semaphore.Interface
+	logger            *zap.Logger
+	sink              SinkI
+	client            http.Client
+	clientMiddleware  func(Client) Client
```
denopink, line 68:
```suggestion
```

let's move this to the webhookv1 and webhookv2 sinks

kafka will use a kafka client and not a http client

sinkSender.go
```diff
@@ -54,17 +48,31 @@ type Sender interface {
...
+	failureMessage    FailureMessage
```
denopink, line 69:
```suggestion
        // failureMessage is sent during a queue overflow.
	failureMessage    FailureMessage
```

denopink (approved):
lgtm!
left some suggestions

PR #478, opened 2024-04-10
updated package structure to include internal package

maurafortino:
Updated package structure to include internal and cmd packages


cmd/main.go
```diff
@@ -0,0 +1,29 @@
...
+
+	"github.com/xmidt-org/caduceus"
+)
+
+func main() {
+	defer func() {
+		if r := recover(); r != nil {
+			fmt.Println("stacktrace from panic: \n" + string(debug.Stack()))
+		}
+	}()
```
denopink, line 19: in terms of logging a panic causing failure, I recommend putting this logic in the fx life cycle 

here's an example of mine: https://github.com/xmidt-org/xmidt-agent/blob/8d8e91fc4e0398c0f0ba864d2cc60ef097772af7/cmd/xmidt-agent/main.go#L203-L268

caduceus.go
```diff
@@ -74,7 +78,7 @@ func caduceus(arguments []string, run bool) error {
 			goschtalt.UnmarshalFunc[sallust.Config]("logging"),
 			goschtalt.UnmarshalFunc[candlelight.Config]("tracing"),
 			goschtalt.UnmarshalFunc[touchstone.Config]("prometheus"),
-			goschtalt.UnmarshalFunc[SinkConfig]("sender"),
+			goschtalt.UnmarshalFunc[sink.SinkConfig]("sender"),
```
denopink, line 81: a general note to help with readability , might be worth bundling fx options in respective package, e.g `sink.Provide()`.

We can always do this later if you want to punt it

caduceus.go
```diff
@@ -124,22 +128,22 @@ func caduceus(arguments []string, run bool) error {
...
 		),
 
-		providePprofEndpoint(),
-		provideMetricEndpoint(),
-		provideHealthCheck(),
-		provideCoreEndpoints(),
+		ProvidePprofEndpoint(),
+		ProvideMetricEndpoint(),
+		ProvideHealthCheck(),
+		ProvideCoreEndpoints(),
```
denopink, line 134: consider lower case

caduceus.go
```diff
@@ -124,22 +128,22 @@ func caduceus(arguments []string, run bool) error {
...
 		arrangehttp.ProvideServer("servers.health"),
 		arrangehttp.ProvideServer("servers.metrics"),
 		arrangehttp.ProvideServer("servers.pprof"),
 		arrangehttp.ProvideServer("servers.primary"),
 		arrangehttp.ProvideServer("servers.alternate"),
 
-		ProvideHandler(),
-		ProvideWrapper(),
+		handler.ProvideHandler(),
+		sink.ProvideWrapper(),
```
denopink, line 143:
```suggestion
		handler.Provide(),
		sink.Provide(),
```

caduceus.go
```diff
@@ -124,22 +128,22 @@ func caduceus(arguments []string, run bool) error {
...
 		touchstone.Provide(),
 		touchhttp.Provide(),
-		ProvideMetrics(),
+		metrics.ProvideMetrics(),
```
denopink, line 146:
```suggestion
		metrics.Provide(),
```

caduceus.go
```diff
@@ -219,7 +223,7 @@ func main() {
 		}
 	}()
 
-	err := caduceus(os.Args[1:], true)
```
denopink, line 222: we want to remove `main` from here since `cmd/main.go` now contains `main`

internal/handler/caduceus_type.go
```diff
@@ -0,0 +1,24 @@
...
+
+type CaduceusHandler struct {
+	wrapper sink.Wrapper
+	Logger  *zap.Logger
+}
+
+func (ch *CaduceusHandler) HandleRequest(workerID int, msg *wrp.Message) {
+	ch.Logger.Info("Worker received a request, now passing to sender", zap.Int("workerId", workerID))
+	ch.wrapper.Queue(msg)
+}
```
denopink, line 24: I strongly recommend to remove middleware like  CaduceusHandler (looks like workerID is always hard coded to `0` and practically not used)

internal/handler/http.go
```diff
@@ -196,11 +197,11 @@ func ProvideHandler() fx.Option {
 		},
 	)
 }
-func New(sw *SinkWrapper, log *zap.Logger, t *HandlerTelemetry, maxOutstanding, incomingQueueDepth int64) (*ServerHandler, error) {
+func New(sw *sink.SinkWrapper, log *zap.Logger, t *HandlerTelemetry, maxOutstanding, incomingQueueDepth int64) (*ServerHandler, error) {
```
denopink, line 200: a general note, let's be consistent with names such as `log *zap.Logger` vs `logger *zap.Logger` if we aren't already

internal/handler/http_test.go
```diff
@@ -88,7 +88,6 @@ func TestServerHandler(t *testing.T) {
 	for _, tc := range tcs {
 		assert := assert.New(t)
 
-		logger := zaptest.NewLogger(t)
```
denopink, line 91: add back test loggers where needed

internal/sink/sinkSender.go
```diff
@@ -94,7 +96,7 @@ type SinkMetrics struct {
 func NewSinkSender(sw *SinkWrapper, l Listener) (sender *SinkSender, err error) {
```
denopink, line 96:
```suggestion
func NewSender(sw *SinkWrapper, l Listener) (sender *Sender, err error) {
```

general note, we should consider renaming things with `sink` as their prefix since now we have the `sink` package

a note about interfaces and structs that may share the same name, assuming there's a need for the given interface, I recommend exposing the interface and privating the struct with the same name


Example:

```go
type Sender interface {
	Update(Listener) error
	Shutdown(bool)
	RetiredSince() time.Time
	Queue(*wrp.Message)
}

type sender struct {
}
```

denopink (approved):
left a few suggestions, otherwise lgtm!

internal/handler/http_test.go
```diff
@@ -1,6 +1,6 @@
 // // SPDX-FileCopyrightText: 2021 Comcast Cable Communications Management, LLC
 // // SPDX-License-Identifier: Apache-2.0
-package main
+package handler
```
denopink, line 3: I recommend to rename any test files packages to `{package_name}_test`, e.g.:
```suggestion
package handler_test
```

PR #480, opened 2024-04-16
Lifecycle

maurafortino:
-Added onStart and onStop lifecycle events to handle panics
-made small re-adjustment to the querylatency metric; what i had updated it to based on the old caduceus updates caused issues with adding labels in various spots

### DIFF
--- a/caduceus.go
+++ b/caduceus.go
@@ -237,7 +237,6 @@ func onStart(logger *zap.Logger) func(context.Context) error {
 			}
 		}()
 
-
 		return nil
 	}
 }
--- a/caduceus.go
+++ b/caduceus.go
@@ -3,8 +3,11 @@
 package caduceus
 
 import (
+	"context"
 	"fmt"
 	"os"
+	"runtime/debug"
+	"time"
 
 	"github.com/alecthomas/kong"
 	"github.com/goschtalt/goschtalt"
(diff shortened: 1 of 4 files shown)


caduceus.go
```diff
@@ -214,3 +226,48 @@ func provideCLIWithOpts(args cliArgs, testOpts bool) (*CLI, error) {
...
+
+	return func(ctx context.Context) error {
+		defer func() {
+			if r := recover(); nil != r {
+				logger.Error("stacktrace from panic", zap.String("stacktrace", string(debug.Stack())), zap.Any("panic", r))
+			}
+		}()
+
+		ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+		defer cancel()
```
denopink, line 241:
```suggestion
```

not used

caduceus.go
```diff
@@ -214,3 +226,48 @@ func provideCLIWithOpts(args cliArgs, testOpts bool) (*CLI, error) {
...
+		return nil
+	}
+}
+
+func onStop(shutdowner fx.Shutdowner, logger *zap.Logger) func(context.Context) error {
+	logger = logger.Named("on_stop")
+
+	return func(_ context.Context) error {
+		defer func() {
+			if r := recover(); nil != r {
```
denopink, line 252:
```suggestion
			if r := recover(); r != nil {
```

PR #483, opened 2024-04-22
Local testing

maurafortino:
made updates that were causing app to fail during local testing


denopink (approved):
lgtm! keep up the great work 🍻

PR #487, opened 2024-04-30
Matcher and Webhook Updates

maurafortino:
- removed the listener version check out of sender.Update
- added NewMatcher function: checks listener version, returns matcher interface, version and error
- added NewSink function: takes in config, version, id, and logger; returns sink interface


internal/sink/matcher.go
```diff
@@ -47,9 +47,24 @@ type CommonWebhook struct {
 	logger *zap.Logger
 }
 
+func NewMatcher(l Listener, logger *zap.Logger) (matcher Matcher, version int, err error) {
```
denopink, line 50: this is more of an artifact from the past, let's take a note to review how loggers are used and passed around in caduceus in general

internal/sink/sink.go
```diff
@@ -39,15 +39,20 @@ type WebhookV1 struct {
...
+func NewSink(c Config, logger *zap.Logger, id string, version int) Sink {
+	var sink Sink
+	if version == 1 {
+		v1 := &WebhookV1{
+			id:               id,
+			deliveryInterval: c.DeliveryInterval,
+			deliveryRetries:  c.DeliveryRetries,
+			logger:           logger,
+		}
+		sink = v1
```
denopink, line 51:
```suggestion
		sink = &WebhookV1{
			id:               id,
			deliveryInterval: c.DeliveryInterval,
			deliveryRetries:  c.DeliveryRetries,
			logger:           logger,
		}
```

internal/sink/sink.go
```diff
@@ -39,15 +39,20 @@ type WebhookV1 struct {
 	// clientMiddleware func(http.Client) http.Client
 }
 
-func NewWebhookV1(s *sender) {
-	v1 := &WebhookV1{
-		id:               s.id,
-		deliveryInterval: s.deliveryInterval,
-		deliveryRetries:  s.deliveryRetries,
-		logger:           s.logger,
+func NewSink(c Config, logger *zap.Logger, id string, version int) Sink {
```
denopink, line 42: based on my other comment https://github.com/xmidt-org/caduceus/pull/487/files#r1586542444
let's add the same `switch` here, otherwise version (and what it means) may not be obvious to anyone other than us
```
switch v := l.(type) {
	case *ListenerV1:
```

denopink (changes requested):
Looks great so far, I just added some suggestions. Keep up the great work!

internal/sink/sink.go
denopink: if we pass the listener then we can use `l.GetId()` instead of passing in the `id`

denopink (approved):
wonderful, this already feels a lot cleaner. Great job and keep it up!

caduceus.yaml
```diff
@@ -291,32 +291,33 @@ authHeader: ["xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx=", "dXNlcjpwYXNz"]
...
+    bucket: "webhooks"
+
+    # address is Argus' network location.
+    address: "http://localhost:6600"
+
+    # auth the authentication method for argus.
+    auth:
+      # basic configures basic authentication for argus.
+      # Must be of form: 'Basic xyz=='
+      basic: "Basic "
```
denopink, line 320: let's change these back as well

internal/anclaHelper/anclaHelper.go
```diff
@@ -0,0 +1,64 @@
...
+	Logger   *zap.Logger
+}
+type AnclaServiceIn struct {
+	fx.In
+	Config   ancla.Config
+	Listener ancla.ListenerConfig
+	Sink     sink.Wrapper
+}
+
+func InitializeAncla(lifecycle fx.Lifecycle) fx.Option {
```
denopink, line 28: lets move this as a `provide.go` file in ancla 🙂

internal/client/httpClient.go
```diff
@@ -5,18 +5,36 @@ package client
...
+}
+
+// httpClientTimeout contains timeouts for an HTTP client and its requests.
+type HttpClientTimeout struct {
+	// ClientTimeout is HTTP Client Timeout.
+	ClientTimeout time.Duration
+
+	// NetDialerTimeout is the net dialer timeout
+	NetDialerTimeout time.Duration
+}
```
denopink, line 37:
```suggestion
}

```

internal/sink/sinkSender.go
```diff
@@ -471,3 +467,25 @@ func (s *sender) CreateMetrics(m metrics.Metrics) {
 	s.maxWorkersGauge = m.ConsumerMaxDeliveryWorkersGauge.With(prometheus.Labels{metrics.UrlLabel: s.id})
 
 }
+
+func getPartnerIds(l ancla.Register) ([]string, error) {
+	switch v := l.(type) {
+	case *ancla.RegistryV1:
+		return v.PartnerIDs, nil
+	case *ancla.RegistryV2:
+		return v.PartnerIds, nil
```
denopink, line 476: maybe we should match the names (partnerids) ... it'll make things simpler in some places

denopink: so let's change ancla's

```go

type RegistryV2 struct {
	PartnerIds   []string
	Registration webhook.RegistrationV2
}
```

to

```go

type RegistryV2 struct {
	PartnerIDs   []string
	Registration webhook.RegistrationV2
}
```

internal/anclaHelper/anclaHelper.go
maurafortino: going to work on in another PR

PR #538, opened 2024-08-27
Feat/kafka

maurafortino:
What's Included:
- added select case for ancla.RegistryV2 for NewSink - only for Kafka 
- added in basic configuration for a Kafka producer

TODO: 
- figure out what configurations will be needed for Kafka producer
- edit Kafka send function
- kafka headers and adding arrays correctly


internal/sink/sink.go
```diff
@@ -213,3 +244,235 @@ func (v1 *WebhookV1) onAttempt(request *http.Request, event string) retry.OnAtte
 
 	}
 }
+
+func (k Kafkas) Update(l ancla.Register) error {
+	return nil
+}
+
+// TODO: probably get rid of urls
```
maurafortino, line 252: copied this from the v1 send - want to discuss

internal/sink/sink.go
```diff
@@ -213,3 +244,235 @@ func (v1 *WebhookV1) onAttempt(request *http.Request, event string) retry.OnAtte
...
+
+		zap.String("Topic", kafkaMsg.Topic),
+		zap.Int32("Partition", partition),
+		zap.Int64("Offset", offset),
+	)
+
+	return nil
+
+}
+
```
maurafortino, line 360: copied all of this from wrphttp addMessageHeaders func - also want to discuss

internal/sink/sink.go
```diff
@@ -213,3 +244,236 @@ func (v1 *WebhookV1) onAttempt(request *http.Request, event string) retry.OnAtte
...
+		contentType = wrp.MimeTypeMsgpack
+		//TODO: do we want to use the wrp encoder or the sarama encoder?
+		buffer := bytes.NewBuffer([]byte{})
+		encoder := wrp.NewEncoder(buffer, wrp.Msgpack)
+		encoder.Encode(msg)
+		body = buffer.Bytes()
+	}
+
+	// Create a new Kafka producer
+	producer, err := sarama.NewSyncProducer(k.brokerAddr, k.config)
```
denopink, line 292: I think we want to cache the producer and not recreate it for every incoming event

internal/sink/sink.go
```diff
@@ -40,6 +41,15 @@ type WebhookV1 struct {
 	// clientMiddleware func(http.Client) http.Client
 }
 
+type Kafkas []*Kafka
+type Kafka struct {
```
denopink, line 45: add a producer field so we can cache the producer once created

internal/sink/sink.go
```diff
@@ -50,6 +60,27 @@ func NewSink(c Config, logger *zap.Logger, listener ancla.Register) Sink {
...
+			}
+
+			//TODO: this is basic set up for now - will need to add more options to config
+			//once we know what we are allowing users to send
+			config := sarama.NewConfig()
+			config.Producer.Return.Successes = true
+			config.Producer.RequiredAcks = sarama.WaitForAll
+			config.Producer.Retry.Max = c.DeliveryRetries //should we be using retryhint for this?
+
+			kafka.config = config
```
denopink, line 80: we can add the producer here
`kafka.producer, err := sarama.NewSyncProducer(kafka.brokerAddr, kafka.config)`

denopink, line 80:
```suggestion
			kafka.config.Producer.Return.Successes = true
			kafka.config.Producer.RequiredAcks = sarama.WaitForAll
			kafka.config.Producer.Retry.Max = c.DeliveryRetries //should we be using retryhint for this?
```

maurafortino: why are we getting rid of NewConfig?

denopink (approved):

PR #541, opened 2024-09-05
Matcher v2 init

maurafortino:
What's Included:
- added MatcherV2 for kafka (would like to refactor the Matchers though as noted previously)
- fixed a couple of formatting issues
- added logger to the kafka struct as it was causing a panic

### DIFF
--- a/internal/sink/matcher.go
+++ b/internal/sink/matcher.go
@@ -43,6 +43,10 @@ type MatcherV1 struct {
 	CommonWebhook
 }
 
+type MatcherV2 struct {
+	matcher map[string]*regexp.Regexp
+	CommonWebhook
+}
 type CommonWebhook struct {
 	mutex  sync.RWMutex
 	logger *zap.Logger
@@ -53,6 +57,13 @@ func NewMatcher(l ancla.Register, logger *zap.Logger) (Matcher, error) {
 	switch v := l.(type) {
 	case *ancla.RegistryV1:
 		m := &MatcherV1{}
+		m.logger = logger
+		if err := m.update(*v); err != nil {
+			return nil, err
+		}
+		return m, nil
+	case *ancla.RegistryV2:
+		m := &MatcherV2{}
 		m.logger = logger
 		if err := m.update(*v); err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


internal/sink/matcher.go
```diff
@@ -194,3 +205,76 @@ func (m1 *MatcherV1) getUrls() (urls *ring.Ring) {
...
+
+	if !matchDevice {
+		m2.logger.Debug("device regex doesn't match", zap.String("event.source", msg.Source))
+		return false
+	}
+	return true
+}
+
+// TODO: this is a big reason why I want to refactor the Matcher logic
+func (m2 *MatcherV2) getUrls() (urls *ring.Ring) {
```
maurafortino, line 278: I don't believe MatcherV2 is going to need getUrls - in my refactored branch i move the urls to the matcher to get rid of this function which is also called here: https://github.com/xmidt-org/caduceus/blob/dad38f9d4e160aedc223f6cac33c67289b341d66/internal/sink/sinkSender.go#L423 and why I originally had getUrls a part of the interface

internal/sink/sink.go
```diff
@@ -68,31 +67,32 @@ func NewSink(c Config, logger *zap.Logger, listener ancla.Register) Sink {
...
 			}
 
+			config := sarama.NewConfig()
 			//TODO: this is basic set up for now - will need to add more options to config
 			//once we know what we are allowing users to send
-			kafka.config.Producer.Return.Successes = true			
-			kafka.config.Producer.RequiredAcks = sarama.WaitForAll
-			kafka.config.Producer.Retry.Max = c.DeliveryRetries //should we be using retryhint for this?
 
+			config.Producer.Return.Successes = true
```
maurafortino, line 78: I added back in the default config because it kept failing due to certain fields not being set. i'm setting these fields this way based on a tutorial I found: https://youtu.be/j6bqJKxb2w0?si=-htt-I0kgF8RqHMy&t=1384

internal/sink/sink.go
```diff
@@ -68,31 +67,32 @@ func NewSink(c Config, logger *zap.Logger, listener ancla.Register) Sink {
...
+			config.Producer.RequiredAcks = sarama.WaitForAll
+			config.Producer.Retry.Max = c.DeliveryRetries //should we be using retryhint for this?
+			kafka.config = config
 			// Create a new Kafka producer
 			producer, err := sarama.NewSyncProducer(kafka.brokerAddr, config)
 			if err != nil {
 				kafka.logger.Error("Could not create Kafka producer", zap.Error(err))
 				return nil
 			}
-			defer producer.Close()
```
maurafortino, line 86: this was moved to here: https://github.com/xmidt-org/caduceus/pull/541/files#diff-06f66c32b314a594bf3f237f16c59ea323846c805e5ae1e2d086e87e3b8edfc3R349

internal/sink/sink.go
```diff
@@ -68,31 +67,32 @@ func NewSink(c Config, logger *zap.Logger, listener ancla.Register) Sink {
 			kafka := &Kafka{
 				id:         l.Registration.CanonicalName,
 				brokerAddr: k.BootstrapServers,
-				topic:      "test",
+				topic:      "quickstart-events",
```
maurafortino, line 70: this is based off the kafka docs for testing locally... i imagine we will be including topic as a part of the kafka webhook struct that users will be sending us?

denopink: yup

PR #543, opened 2024-09-09
Matcher refactor

maurafortino:
What's Included:
-Refactored the Matcher functionality so that it was only checking the matcher field for v2 and the matcher and events fields for v1
-Moved the urls functionality out of Matcher into Sink

Future TODO: renaming the Sink structs so as not to get confused with the webhook-schema structs


internal/sink/matcher.go
```diff
@@ -30,26 +26,22 @@ func (c *ClientMock) Do(req *http.Request) (*http.Response, error) {
 // move to subpackage and change to Interface
 type Matcher interface {
 	IsMatch(*wrp.Message) bool
-
```
maurafortino, line 33: removed getUrls from Matcher interface. getUrls is now being called in sink.go

internal/sink/matcher.go
```diff
@@ -30,26 +26,22 @@ func (c *ClientMock) Do(req *http.Request) (*http.Response, error) {
...
-	//TODO: not sure if this will be functionality of all webhooks or just v1
-	//leaving for now - will make changes if running into roadblock with this
-	getUrls() *ring.Ring
 }
 
 type MatcherV1 struct {
 	events  []*regexp.Regexp
 	matcher []*regexp.Regexp
-	urls    *ring.Ring
```
maurafortino, line 42: moved urls to sink.go (webhookV1)

maurafortino: any functionality related to urls that was in matcher.go has also been moved to sink.go

internal/sink/sink.go
```diff
@@ -27,84 +30,141 @@ import (
 )
 
 type Sink interface {
-	Update(ancla.Register) error
```
maurafortino, line 30: still have the update function - but the signature has been updated and is no longer required for the Sink interface

internal/sink/sink.go
```diff
@@ -27,84 +30,141 @@ import (
...
-	deliveryRetries  int
-	logger           *zap.Logger
+	urls *ring.Ring
+	CommonWebhook
 	//TODO: need to determine best way to add client and client middleware to WebhooV1
 	// client           http.Client
 	// clientMiddleware func(http.Client) http.Client
 }
 
+type Webhooks []*WebhookV1
```
maurafortino, line 44: going to use this webhookV1 struct (will be renaming this so as not to be confused with the webhookV1 schema)  as the struct for the newest webhook struct in webhook schema for now - might have to add a third struct to satisfy all registrations

internal/sink/sink.go
```diff
@@ -27,84 +30,141 @@ import (
...
+	id               string
+	failureUrl       string
+	deliveryInterval time.Duration
+	deliveryRetries  int
+	mutex            sync.RWMutex
+	logger           *zap.Logger
+}
 type Kafkas []*Kafka
 type Kafka struct {
-	id         string
```
maurafortino, line 46: id and logger are both in comonwebhook struct

internal/sink/sink.go
```diff
@@ -27,84 +30,141 @@ import (
...
-				topic:      "quickstart-events",
-				logger:     logger,
+		if len(l.Registration.Webhooks) > 0 {
+			var whs Webhooks
+			for _, wh := range l.Registration.Webhooks {
+				v1 := &WebhookV1{}
+				v1.Update(c, logger, wh.ReceiverURLs[1:], l.GetId(), l.Registration.FailureURL, wh.ReceiverURLs[0])
+				whs = append(whs, v1)
 			}
-
```
maurafortino, line 73: all of this has been move to the kafka.Update function

internal/sink/sink.go
```diff
@@ -27,84 +30,141 @@ import (
...
+	urlCount, err := getUrls(altUrls)
+	if err != nil {
+		l.Error("error recevied parsing urls", zap.Error(err))
+	}
+	v1.updateUrls(urlCount, receiverUrl, altUrls)
+
 	return nil
 }
 
+func (v1 *WebhookV1) updateUrls(urlCount int, url string, urls []string) {
```
maurafortino, line 111: moved over this functionality from mathcer.go

internal/sink/sinkSender.go
```diff
@@ -420,7 +418,6 @@ Loop:
 			}
 			s.queueDepthGauge.Add(-1.0)
 			s.mutex.RLock()
-			urls = s.matcher.getUrls()
```
maurafortino, line 423: urls is now added to the sink structs and we will be referencing them directly from the struck rather than sending them to the send function

PR #556, opened 2024-10-04
Hashing Logic

maurafortino:
What's Included:
- added docker-compose file for running multiple kafka brokers
- added hashing feature


docker-compose.yml
```diff
@@ -0,0 +1,143 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+version: '2'
+services:
+  controller-1:
+    image: ${IMAGE}
```
denopink, line 20: recommend fixing it to kafka's latest for now

denopink: maybe we should hold off on committing this file as well. there's probably a better place to commit this

internal/sink/hasher.go
```diff
@@ -0,0 +1,81 @@
...
+	return len(h)
+}
+func (h HashRing) Less(i, j int) bool {
+	return h[i].hash < h[j].hash
+}
+func (h HashRing) Swap(i, j int) {
+	h[i], h[j] = h[j], h[i]
+}
+
+func (h HashRing) GetServer(key string) string {
```
denopink, line 31: I think you can get rid of `Server` in these names

internal/sink/hasher.go
```diff
@@ -0,0 +1,81 @@
...
+			*h = append((*h)[:i], (*h)[i+1:]...)
+			break
+		}
+	}
+	sort.Sort(h)
+}
+
+func NewRing() *HashRing {
+	return &HashRing{}
+}
```
denopink, line 65: doesn't add much value

```suggestion
```

internal/sink/sink.go
```diff
@@ -76,12 +80,19 @@ func NewSink(c Config, logger *zap.Logger, listener ancla.Register) Sink {
 			return whs
 		}
 		if len(l.Registration.Kafkas) > 0 {
-			var sink Kafkas
-			for _, k := range l.Registration.Kafkas {
+			var sink KafkaSink
+			r := NewRing()
```
denopink, line 84:
```suggestion
			r := HashRing{}
```

internal/sink/sink.go
```diff
@@ -76,12 +80,19 @@ func NewSink(c Config, logger *zap.Logger, listener ancla.Register) Sink {
...
 				kafka := &Kafka{}
 				kafka.Update(l.GetId(), "quickstart-events", k.RetryHint.MaxRetry, k.BootstrapServers, logger)
-				sink = append(sink, kafka)
+				sink.Kafkas = append(sink.Kafkas, kafka)
+				if l.Registration.Hash.Field != "" {
+					key := l.Registration.Hash.Field + strconv.Itoa(i)
+					r.AddServer(key)
+				}
 			}
+			sink.Hash = r
```
denopink, line 95:
```suggestion
			sink.Hash = &r
```

internal/sink/sink.go
```diff
@@ -91,8 +102,7 @@ func NewSink(c Config, logger *zap.Logger, listener ancla.Register) Sink {
 }
 
 func (v1 *WebhookV1) Update(c Config, l *zap.Logger, altUrls []string, id, failureUrl, receiverUrl string) (err error) {
-	//TODO: is there anything else that needs to be done for this?
-	//do we need to return an error
+	//TODO: do we need to return an error if not - we should get rid of the error return
```
denopink, line 105: if we don't need it, then we need to see whether the error fulfills any interfaces that we depend on

internal/sink/sink.go
```diff
@@ -338,14 +348,29 @@ func (k *Kafka) Update(id, topic string, retries int, servers []string, logger *
...
-	//for now: when no hash is given we will just loop through all the kafkas
+func (k KafkaSink) Send(secret string, acceptType string, msg *wrp.Message) error {
 	var errs error
-	for _, kafka := range k {
-		err := kafka.send(secret, acceptType, msg)
-		if err != nil {
-			errs = errors.Join(errs, err)
+	if k.HashField != "" {
+		key := GetKey(k.HashField, msg)
+		for i, kafka := range k.Kafkas {
```
denopink, line 355: I like what you did here, but I think we can make this easier to maintain and read 🙂 

tl;dr i would suggest to leverage the fact you're using hashes by converting `k.Kafkas` to a map and then using `k.Hash.GetServer(key)` directly, e.g.: `err := k.Kafkas[GetKey(k.HashField, msg)].send(secret, acceptType, msg)`

that would simplify the code a bit

denopink (changes requested):
great work as always 💯 

I have a few suggestions 🙂

.gitignore
```diff
@@ -51,9 +51,10 @@ _testmain.go
 .vscode/*
 .dev/*
 
-caduceus
```
maurafortino, line 54: Wes had me move the main.go file from cmd/main.go to cmd/caduceus/main.go for correct functionality of goschtalt - this line caused the main.go file to be ignored so added yaml so it would ignore changes to just the yaml file

docker-compose.yml
maurafortino: yeah i'm good to remove this file for now - i did not intend to push it i just forgot to remove it when i ran `git add .`

internal/sink/sink.go
```diff
@@ -338,14 +348,28 @@ func (k *Kafka) Update(id, topic string, retries int, servers []string, logger *
...
+	if k.HashField != "" {
+
+		if kafka, ok := k.Kafkas[k.Hash.Get(GetKey(k.HashField, msg))]; ok {
+			err := kafka.send(secret, acceptType, msg)
+			if err != nil {
+				errs = errors.Join(errs, err)
+			}
+
+		} else {
+			errs = fmt.Errorf("could not find kakfa for the related hash %v", k.HashField)
```
maurafortino, line 362: do we want to send an error here? or instead of the k.HashField if check do we want say if err != nil we go straight to the for loop of allt he kafkas?

denopink: lets leave this, but we need to think about how to best describe possible error cases like this, i.e. define errors, new metrics, logging (I think this section is good, just add a todo saying we need to flush out the error handling for kafka)

denopink: technically hashing can't fell as long as there is 1 item so `k.Kafkas[k.Hash.Get(GetKey(k.HashField, msg))]` should always be good

but to make sure of this, we can add a check `len(k.Kafkas) == len(k.Hash)` during the Hash and kafka setup

internal/sink/sink.go
```diff
@@ -338,14 +350,24 @@ func (k *Kafka) Update(id, topic string, retries int, servers []string, logger *
...
-func (k Kafkas) Send(secret string, acceptType string, msg *wrp.Message) error {
-	//TODO: discuss with wes and john the default hashing logic
-	//for now: when no hash is given we will just loop through all the kafkas
+func (k KafkaSink) Send(secret string, acceptType string, msg *wrp.Message) error {
 	var errs error
-	for _, kafka := range k {
-		err := kafka.send(secret, acceptType, msg)
-		if err != nil {
-			errs = errors.Join(errs, err)
+	if len(*k.Hash) == len(k.Kafkas) {
```
denopink, line 355: I think we want to the do this check differently, but we're ok to merge this for now and I'll patch it in a new pr

PR #587, opened 2024-12-11
added logic for the v2 webhooks including srv lookup and hashing

maurafortino:
What's Included:
- Updated logic for the V2 webhooks including:
- hashing logic
- `custom dialer` and `dial context` func for using the SRVs returned in new `lookup` function

### DIFF
--- a/internal/sink/sink.go
+++ b/internal/sink/sink.go
@@ -5,6 +5,7 @@ package sink
 import (
 	"bytes"
 	"container/ring"
+	"context"
 	"crypto/hmac"
 	"crypto/sha1"
 	"encoding/hex"
@@ -12,8 +13,10 @@ import (
 	"fmt"
 	"io"
 	"math/rand"
+	"net"
 	"net/http"
 	"net/url"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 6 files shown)


internal/sink/sink.go
```diff
@@ -71,12 +85,43 @@ func NewSink(c Config, logger *zap.Logger, listener ancla.Register) Sink {
...
+					v2.updateUrls(urlCount, wh.ReceiverURLs[0], wh.ReceiverURLs[1:])
+				} else {
+					sortBy := wh.DNSSrvRecord.LoadBalancingScheme
+					for _, domain := range wh.DNSSrvRecord.FQDNs {
+						dialer := new(Dialer)
+						err := dialer.lookup(domain, sortBy)
+						if err != nil {
+							v2.logger.Error("error received looking up service records", zap.Error(err))
+						}
+						dialer.NewClient()
```
maurafortino, line 115: do we want to create the client here? or does it make more sense to move to the `send()` function?

denopink: Here is good, I think I'll push up some updates to help you with the structure. (I'll slack you)

denopink (commented):
This is a great start! 🙂  I'll push up some suggested changes. (I'll slack you)

PR #594, opened 2025-01-09
Dns package

maurafortino:
What's Included:
- moved srvRecordDialer to its own package (dns)
- removed any instances of non standard go libraries from dns package
- created tests for NewSRVRecordDialer

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/xmidt-org/retry v0.0.3
 	github.com/xmidt-org/sallust v0.2.2
 	github.com/xmidt-org/touchstone v0.1.5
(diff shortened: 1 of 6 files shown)


internal/dns/srvRecordDialer.go
```diff
@@ -0,0 +1,86 @@
...
+	}, nil
+
+}
+
+type SRVRecordDialer struct {
+	srvs  []*net.SRV
+	fqdns []string
+}
+
+func (d *SRVRecordDialer) DialContext(ctx context.Context, _, _ string) (net.Conn, error) {
```
maurafortino, line 70: i think this needs to be rewritten to properly test. Do we want to be hitting the net.Dial function when we test?

maurafortino:
Closing and creating new PR to fix large merge request

PR #595, opened 2025-01-09
Dns package

maurafortino:
What's Included:
- moved SRVRecordDialer to its own packaged
- removed any instances of nonstandard go libraries from package
- added "load balancing" for weighted SRV records
- created tests for NewSRVRecordDialer


internal/dns/srvRecordDialer_test.go
```diff
@@ -0,0 +1,184 @@
...
+	"context"
+	"net"
+	"net/http"
+	"testing"
+
+	"github.com/foxcpp/go-mockdns"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewSRVRecordDialer(t *testing.T) {
```
johnabass, line 14: Consider using `github.com/stretchr/testify/suite` for tests. That package lets you organize your tests better, moving setup/teardown code into separate methods, etc. It provides a lot of the same structure as frameworks like TestNG for Java.

internal/dns/srvRecordDialer.go
```diff
@@ -0,0 +1,136 @@
...
+	"net"
+	"net/http"
+	"sort"
+)
+
+type Resolver interface {
+	LookupSRV(ctx context.Context, service, proto, name string) (string, []*net.SRV, error)
+}
+
+func NewSRVRecordDialer(fqdns []string, sortBy string, resolver Resolver) (http.RoundTripper, error) {
```
johnabass, line 17: You probably want to use the functional options pattern here. This function is a primary entry point for this package and it's signature will likely need to change as this package evolves over time. Using functional options allows clients to be backwards compatible and resilient in the face of new features.

internal/dns/srvRecordDialer.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (d *SRVRecordDialer) DialContext(ctx context.Context, _, _ string) (net.Conn, error) {
+	var errs error
+	var err error
+	var conn net.Conn
+
+	//TODO: add retry logic if we receive conn error? or just move to next one?
+	switch d.sortBy {
+	case "weight":
```
johnabass, line 70: Instead of using constants whose value you switch on, a more extensible way would be to create a `Sorter` strategy that did the sorting.  Then, instead of passing `sortBy`, you'd have someone pass this code a `Sorter`. This package can provide (2) implementations of `Sorter`: one for weight and one for priority.

denopink: I think this sorting logic is going to be replace soon with actual sampling/load-balancing code, right @maurafortino ?

internal/dns/srvRecordDialer.go
```diff
@@ -0,0 +1,136 @@
...
+			if err != nil {
+				errs = errors.Join(errs,
+					fmt.Errorf("%v: host `%s` [weight: %d, priortiy: %d] from srv record `%v`",
+						err, host, addr.Weight, addr.Priority, d.fqdns))
+				continue
+			}
+			return conn, errs
+		}
+	default:
+		return nil, fmt.Errorf("unknown loadBalancingScheme type: %s", d.sortBy)
```
denopink, line 106: best to convert this to a package error for testing and general error handling

internal/dns/srvRecordDialer.go
```diff
@@ -0,0 +1,136 @@
...
+		sortBy: sortBy,
+	}
+
+	var errs error
+
+	for _, fqdn := range d.fqdns {
+		_, addrs, err := resolver.LookupSRV(context.Background(), "", "", fqdn)
+		if err != nil {
+			errs = errors.Join(errs,
+				fmt.Errorf("srv lookup failure: `%s`", fqdn),
```
denopink, line 37: best to convert this to a package error for testing and general error handling

internal/dns/srvRecordDialer.go
```diff
@@ -0,0 +1,136 @@
...
+			)
+			continue
+		}
+
+		d.srvs = append(d.srvs, addrs...)
+	}
+
+	// TODO: ask wes/john whether 1 or more net.LookupSRV error should trigger an error from NewSRVRecordDailer
+	if len(d.srvs) == 0 {
+		return nil, errors.Join(fmt.Errorf("expected atleast 1 srv record from fqdn list `%v`", d.fqdns), errs)
```
denopink, line 48: best to convert this to a package error for testing and general error handling

denopink (changes requested):
Left a few suggestions, great progress so far! 🍻

internal/dns/srvRecordDialer.go
```diff
@@ -0,0 +1,136 @@
+package dns
```
denopink, line 1: add copyright

internal/dns/srvRecordDialer_test.go
```diff
@@ -0,0 +1,184 @@
+package dns
```
denopink, line 1: add copyright

denopink (changes requested):
more suggestions! 😄

internal/dns/srvRecordDialer.go
maurafortino: @denopink   there is sampling/load-balancing code in the getAddrByWeight function nested inside the switch statement: https://github.com/xmidt-org/caduceus/blob/4a72c1e544ab1c9c609872b1f6624bd92269f4e5/internal/dns/srvRecordDialer.go#L111C36-L112C21

maurafortino: Spoke with John and we're going to be changing from sorter strategy to chooser strategy

internal/dns/srvRecordDialer.go
maurafortino: can you clarify what you mean by package error?

internal/dns/srvRecordDialer.go
```diff
@@ -0,0 +1,136 @@
...
+	var errs error
+	var err error
+	var conn net.Conn
+
+	//TODO: add retry logic if we receive conn error? or just move to next one?
+	switch d.sortBy {
+	case "weight":
+		//create a copy of d.srvs so that we can edit the srvs list if needed
+		srvs := make([]*net.SRV, len(d.srvs))
+		copy(srvs, d.srvs)
```
denopink, line 73: why are we making a copy?

internal/dns/srvRecordDialer.go
denopink: package level variable with an assigned error

maurafortino: is that common practice for us? i don't think i've ever seen us make error a package level variable. do you have an example?  maybe i'm not understanding fully.

internal/dns/srvRecordDialer.go
maurafortino: > I may have missed the conversation behind the algo being used in `getAddrByWeight`.
> 
> I recommend adding a short description of the expected behavior of `getAddrByWeight`.

@denopink I don't think we had a discussion on which algo to use. i'm just using weighted random for now. i can set up a time though to discuss if needed.

internal/dns/srvRecordDialer.go
denopink: https://github.com/xmidt-org/ancla/blob/cf3eaffa4279a79ce69f41da9432e5266dbe1507/chrysom/basicClient.go#L27-L36

internal/dns/srvRecordDialer.go
denopink: got it, just letting you know I don't think getAddrByWeight is giving us “weighted random” load balancing. weighted random load balancing usually means routing requests evenly across healthy targets in a random order.

No need for a meeting (imo) if you aren't focusing on the load balancing specifics just yet, I just wanted to point it out in case I was missing something. Sounds like it's a place holder while you're working on this. thx for clarifying 🙂

internal/dns/srvRecordDialer.go
maurafortino: oooh yeah was def not understanding fully. thx!

internal/sink/srvRecordDialer.go
maurafortino, line 1: moved to internal/dns/srvRecordDialer.go

internal/dns/chooser.go
maurafortino, line 1: @denopink @schmidtw  I would like discuss how we are going to be choosing between weight and priority. @johnabass mentioned we normally would choose one or the other (he wasn't sure exactly which one is chosen but for this example we'll go with weight) and then if there are two with the same weight we look to see which has the higher priority. And then vice versa if we use priority. Is this how we want to do that or are we thinking of distributing events differently?

maurafortino:
pausing work on this for now after discussing with John and Wes. Will work on in the background - but prioritizing batching and releasing new caduceus before finalizing the design for "load balancing" the SRVs

maurafortino:
closing this PR for now - could be reopened when we revisit the SRV logic

PR #596, opened 2025-01-31
Dns package rewrite

maurafortino:
What's Included
-updated the logic of the chooser so we only have a priority chooser
-priority chooser is sorted in ascending order
-map is created so that if there are mutliple srvs with the same priority they are grouped together
-we then choose an srv from those srvs based on their weight
Question: is this current logic even correct for a round robin choice? 
Question: if correct: if there are multiple srvs for a priority and the one chosen fails do we want to try the other weights or do we just want to skip ahead to the next priority


maurafortino:
Closing this PR for now - could be reopened when we revisit the SRV logic

PR #600, opened 2025-03-12
batching edits for batchMsgs and sendBatch

maurafortino:
- updated the batching logic in batchMsgs and sendBatch
- added worker function

### DIFF
--- a/internal/sink/sink.go
+++ b/internal/sink/sink.go
@@ -29,6 +29,7 @@ import (
 	"go.uber.org/zap"
 )
 
+// TODO: update the structure of this function - remove error (and possibly remove the strings)
 type Sink interface {
 	Send(string, string, *wrp.Message) error
 }
@@ -43,6 +44,7 @@ type WebhookV1 struct {
 
 type WebhookV2 struct {
 	urls *ring.Ring
+	ch   chan struct{}
 	CommonWebhook
 	//TODO: need to determine best way to add client and client middleware to WebhooV1
 	// clientMiddleware func(http.Client) http.Client
@@ -54,6 +56,7 @@ type WebhookSink struct {
 	BatchMessages int
 	BatchLinger   time.Duration
 	ch            chan *wrp.Message
+	workers       int
 }
 
 // TODO: rename this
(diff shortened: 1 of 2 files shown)


denopink (changes requested):
We had a review over a slack huddle and the tl;dr is that I pointed maura to xmidt-agent's qos as a guide 

https://github.com/xmidt-org/xmidt-agent/blob/main/internal/wrphandlers/qos/qos.go

internal/sink/sink.go
```diff
@@ -225,43 +229,65 @@ func (whs WebhookSink) Send(secret, acceptType string, msg *wrp.Message) error {
...
+    2a. identify potential resource starvation (i.e. running too many threads than CPU cores)
+    2b. use worker interface (aquire & release) to enforce rate limiting
+    2b1. where should worker.Acquire be called and where should worker.Release be called
+
+3. how do we define sendBatch to be ready. what conditions signal channel to be closed/is ready?
+*/
+
+func (whs WebhookSink) sendBatch(msgs []*wrp.Message) <-chan struct{} {
+	ready := make(chan struct{})
+	defer close(ready)
```
denopink, line 267: we need to add some sort of condition that determines when to call `close(ready)`. Currently, exiting `sendBatch` triggers it and that's not really doing thing useful.

denopink (changes requested):
The call to `close(ready)` requires a trigger/condition otherwise it's not very useful as is.

Keep up the good work!

internal/sink/sink.go
```diff
@@ -745,12 +754,13 @@ func (v2 *WebhookV2) updateUrls(urlCount int, url string, urls []string) {
...
 // them to the listeners outside webpa
-func (v2 *WebhookV2) send(msg *wrp.Message) error {
+func (v2 *WebhookV2) send(msg *wrp.Message, wg *sync.WaitGroup) error {
 	defer func() {
 		if r := recover(); nil != r {
 			// s.DropsDueToPanic.With(prometheus.Labels{metrics.UrlLabel: s.id}).Add(1.0)
 			v2.logger.Error("goroutine send() panicked", zap.String("id", v2.id), zap.Any("panic", r))
 		}
+		wg.Done()
 		// s.workers.Release()
```
denopink, line 764: Instead of creating `wg *sync.WaitGroup` for every `batchSend`, we could leverage something like `WebhookV2`'s `s.workers.Release()`. But it'll live in `WebhookSink` instead. This would act as a rate limiter, making sure 1 Sink is not using all the server's resources. It doesn't have to be a semaphore, a channel can provide the same functionality

internal/sink/sink.go
```diff
@@ -123,6 +130,8 @@ func NewSink(c Config, logger *zap.Logger, listener ancla.Register) Sink {
 				}
 			}
 
+			whs.lock.Lock() //QUESTION: would this go here or inside the goroutine? if here do we need to change signature of this function?
+			defer whs.lock.Unlock()
```
denopink, line 134: We'll incorporate a lock soon, it'll be used during when the sink is either updated, started or shutdown

internal/sink/sink.go
```diff
@@ -222,46 +231,59 @@ func (whs WebhookSink) Send(secret, acceptType string, msg *wrp.Message) error {
...
 	ticker := time.NewTicker(whs.BatchLinger) //TODO: setting to this for now - open to suggestions
-	select {
-	case msg := <-whs.ch:
-		msgs = append(msgs, msg)
-		if len(msgs) == whs.BatchMessages {
-			whs.sendBatch(msgs)
+	expired := false
+	for {
+		select {
+		case <-whs.readyCh:
```
denopink, line 240: we should think about the readyCh as something to be used and thrown away (replaced), it's not something to store as state

internal/sink/sink.go
```diff
@@ -222,46 +231,59 @@ func (whs WebhookSink) Send(secret, acceptType string, msg *wrp.Message) error {
...
+
+3. how do we define sendBatch to be ready. what conditions signal channel to be closed/is ready?
+*/
+
+func (whs *WebhookSink) sendBatch(msgs []*wrp.Message, eventSink WebhookV2) chan struct{} {
+	//assuming that messages all point to the same webhook hash
+	var done chan struct{}
+	go func(msg *wrp.Message) { //TODO: this will change when we add in content-type logic
+		defer close(done)
+		done <- struct{}{}
```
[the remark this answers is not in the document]
denopink: we would close this channel after the rate limiter logic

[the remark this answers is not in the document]
denopink: i recommend, either calling both channels `ready` or `done` but not a mix (in case it confuses future readers).

internal/sink/sink.go
```diff
@@ -719,12 +741,13 @@ func (v2 *WebhookV2) updateUrls(urlCount int, url string, urls []string) {
...
 // worker is the routine that actually takes the queued messages and delivers
 // them to the listeners outside webpa
-func (v2 *WebhookV2) send(secret, acceptType string, msg *wrp.Message) error {
+func (v2 *WebhookV2) send(msg *wrp.Message, ch chan struct{}) error { //TODO: this will change when we add in content-type logic
 	defer func() {
 		if r := recover(); nil != r {
 			// s.DropsDueToPanic.With(prometheus.Labels{metrics.UrlLabel: s.id}).Add(1.0)
 			v2.logger.Error("goroutine send() panicked", zap.String("id", v2.id), zap.Any("panic", r))
 		}
+		<-ch
```
denopink, line 750: this ch has the potential of causing chaos and it's not aiding to the cause (even tho it may feel like it is)

denopink (changes requested):
Keep up the great work! 🍻

internal/sink/sink.go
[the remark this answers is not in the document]
denopink: @maurafortino you missed this suggestion

denopink (changes requested):
looks good so far, don't forget to add the limiter logic https://github.com/xmidt-org/caduceus/pull/600#discussion_r2023185182

denopink:
closing for now since we're taking care of batching after refactoring the `sink` package

PR #604, opened 2025-04-10
Handler tests

maurafortino:
- created new caduceusHandler that includes the sink.Wrapper and a logger to fix unit tests
- updated the serverHandler metrics field to reflect the struct that includes only the metrics needed for the serverHandler
- added a mock_test file for the requestHandler mock
- moved the metrics mocks to the mock.go file under the metrics folder


internal/handler/http.go
```diff
@@ -26,19 +26,32 @@ type ServerHandlerIn struct {
...
-	Telemetry   *Telemetry
+	Metrics     metrics.ServerHandlerMetrics
 }
 
 type ServerHandlerOut struct {
 	fx.Out
 	Handler *ServerHandler
 }
 
+type RequestHandler interface {
```
maurafortino, line 37: this was added to better implement unit tests - only needed to mock the handleRequest function and not the ServeHTTP function. this allows us to mock handleRequest without also having to mock ServeHTTP

internal/handler/http_test.go
maurafortino, line 1: tests were commented back in - all pass. only changes made were to update variable and field names.

internal/mocks/mocks.go
maurafortino, line 1: was having issues with circular dependencies so split mocks into their own files under the appropriate folders

internal/mocks/mocks.go
```diff
@@ -1,82 +0,0 @@
...
-	MaxOutstanding     int64
-	Now                func() time.Time
-}
-
-func (m *Handler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
-	m.Called(r)
-}
-
-// mockSenderWrapper needs to mock things that the `SenderWrapper` does
-type Wrapper struct {
```
maurafortino, line 35: removed Wrapper struct entirely as of right now it's not being used. Will most likely end up adding it back in for future tests

denopink:
I drafted this pr since it's not ready. @maurafortino will undraft when it's ready

denopink (approved):
Ran the tests locally and they work as expected! There may be a few things we want to clean up from the old tests, but those can wait, lgtm! 🍻

